PR #3, opened 2017-09-06
Make missing CLICON_BACKEND_DIR message match the key name.

twohundredio:
Fix up the wrong key name in error message when CLICON_BACKEND_DIR is missing in the config file.


olofhagsand:
Thanks! 
I'll merge it into the develop branch, will appear in master eventually.

PR #4, opened 2017-12-04
Fix make distclean without keyvalue

rbgarga:
When --with-keyvalue is not selected in configure, keyvalue is not added
to SUBDIRS variable under datastore Makefile and when distclean runs
datastore/keyvalue/Makefile is left behind. Add keyvalue to the SUBDIRS
loop using sort function to prevent it to be duplicated


rbgarga:
Thanks!

PR #12, opened 2018-01-31
Make default config to respect sysconfdir

rbgarga:
Instead of hardcode default config to /etc/clixon.xml, make it to
respect sysconfdir variable. Since autoconf doesn't expand variables on
header substitution it would end up being defined on clixon_config.h as
"${prefix}/etc/clixon.xml" what makes no sense for the header file.

To workaround this no-expand-variable issue I renamed clixon_config.h.in
to clixon_config.h.in.in and defined default value of
CLIXON_DEFAULT_CONFIG as %%sysconfdir%%/clixon.xml. Then
include/Makefile will be responsible to run the 2nd pass substitution
and using sed replace %%sysconfdir%% by expanded value of ${sysconfdir}.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ etc/clixonrc
 example/*.conf
 
 include/clixon_config.h
+include/clixon_config.h.in
 
 lib/src/build.c
 lib/clixon/clixon.h
--- a/configure
+++ b/configure
@@ -2161,7 +2161,7 @@ else
 	CLIGEN_PREFIX="$prefix"
 fi
 
-ac_config_headers="$ac_config_headers include/clixon_config.h lib/clixon/clixon.h"
+ac_config_headers="$ac_config_headers include/clixon_config.h.in lib/clixon/clixon.h"
 
 
 
@@ -4053,7 +4053,7 @@ fi
 if test "${with_configfile+set}" = set; then :
   withval=$with_configfile; DEFAULT_CONFIG="$withval"
 else
-  DEFAULT_CONFIG="/etc/clixon.xml"
+  DEFAULT_CONFIG="%%sysconfdir%%/clixon.xml"
 fi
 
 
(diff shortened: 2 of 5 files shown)


olofhagsand:
It seems too non-main-line. I understand the usecase but wouldlike to stay within normal usage if possible. 
What about:
olof@vandal> git diff configure.ac
diff --git a/configure.ac b/configure.ac
index 941f8e3..5181728 100644
--- a/configure.ac
+++ b/configure.ac
@@ -165,7 +165,7 @@ fi
 AC_ARG_WITH([configfile],
            [AS_HELP_STRING([--with-configfile=FILE],[set default path to config file])],
            [DEFAULT_CONFIG="$withval"],
-           [DEFAULT_CONFIG="/etc/clixon.xml"])
+           [DEFAULT_CONFIG="$prefix/etc/clixon.xml"])
 
 AC_CHECK_LIB(crypt, crypt)
 AC_CHECK_HEADERS(crypt.h)

The drawback is thatit does not honor a separate setting of sysconfdir, you would have to set --with-prefix. Would that be good enough?

rbgarga:
Format is making it hard to read your comment. Do you mind to edit it and add the git diff block under ```?

rbgarga:
The biggest problem is the way autoconf was designed to work, keeping main variables unexpanded to make it possible for users to replace them during `make` phase. It works pretty well for Makefiles and for shell scripts but fails for .h files. This double parse workaround is the way used by many projects to bypass autoconf poor design.

Since --sysconfdir is the configure knob designed to define where config files must be installed IMO it would be correct to respect that. I say that not as a developer but as a package maintainer. I've been working as a committer of FreeBSD ports collection for many years and for packagers it's hard to deal with projects that doesn't respect standards.

If we just change the default to add `$prefix` before `/etc/` it will end up being the same with a different default value so I believe the change would be unnecessary.

I have an idea to make it work in a simpler form and will submit a new PR soon.

PR #13, opened 2018-02-06
Make default config to respect sysconfdir

rbgarga:
Instead of hardcode default config to /etc/clixon.xml, make it to
respect sysconfdir variable. Since autoconf doesn't expand variables on
header substitution it would end up being defined on clixon_config.h as
"${prefix}/etc/clixon.xml" what makes no sense for the header file.

Use eval to expand ${sysconfdir} and make .h file to be defined with
final directory name.

### DIFF
--- a/configure
+++ b/configure
@@ -4053,7 +4053,7 @@ fi
 if test "${with_configfile+set}" = set; then :
   withval=$with_configfile; DEFAULT_CONFIG="$withval"
 else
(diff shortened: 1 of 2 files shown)


rbgarga:
@olofhagsand This solution is much simpler and end up solving the case

olofhagsand:
Nice solution, thanks. Except "tnsr" should be "clixon".

rbgarga:
Ooops. Fixed now. Thanks!

PR #19, opened 2018-04-23
Improve permissions and strip binaries

rbgarga:
- Use 0755 for directories
- Use 0644 for libraries, includes and shared files
- Use -s (strip) parameter when installing binaries and libraries


olofhagsand:
Actually -s caused problems for me running debugging mode, eg CFLAGS=-g ./configure. I wish to be able to have the symbol table while debugging. Any suggestions on how to do this? Define a INSTALLFLAGS in configure.ac?

rbgarga:
There are several ways to deal with that. One is to use a knob to define if we are building with debug or not and then check for it and define macros for install binaries and libraries like
```
INSTALL_PROGRAM=install -m 0755 -s
INSTALL_LIBRARY=install -m 0644 -s
```
when DEBUG is not set and when we do `make DEBUG=1` it removed `-s` from it

rbgarga:
or even create a `./configure --with-debug-symbols` or whatever parameter is standard for autotools

PR #27, opened 2018-06-07
add handling of CDATA to XML parser

dcornejo:
This add CDATA handling to clixon.

Anything data wrapped in "<![CDATA" ... "]]>" will be passed through transparently w/o further scrutiny. I don't see any case where this affects existing working code.

This has been tested in my use, any changes, clarification, or commentary welcome.

### DIFF
--- a/lib/src/clixon_xml_parse.l
+++ b/lib/src/clixon_xml_parse.l
@@ -118,6 +118,7 @@ int clixon_xml_parsewrap(void)
 <AMPERSAND>"aquot; "   {BEGIN(_YA->ya_lex_state); clixon_xml_parselval.string = "'"; return CHARDATA;}
 
 <CDATA>.              { clixon_xml_parselval.string = yytext; return CHARDATA;}
+<CDATA>\n            { clixon_xml_parselval.string = yytext;_YA->ya_linenum++; return (CHARDATA);}
 <CDATA>"]]>"          { BEGIN(_YA->ya_lex_state); clixon_xml_parselval.string = yytext; return CHARDATA;}
 
 <CMNT>"-->"           { BEGIN(START); return ECOMMENT; }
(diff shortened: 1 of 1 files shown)


dcornejo:
I didn't handle newlines properly, hence another commit, sorry...

olofhagsand:
@dcornejo Just for clarification. 
I see the code incorporates the CDATA including the "meta-syntax" so that for example:
  <a><![CDATA[hello&]]></a>
is internally represented as:
element a {
    body value:"<![CDATA[hello&j]]>"
}
which means that when you print it, or save it in the datastore it will keep the CDATA syntax.
I would (kind of) have thought you wanted to strip the "meta-syntax" so it translated to eg:
element a {
    body value:"hello&"
}
which is the case with encode/decode.
If this is OK I am happy, just wanted to double check?
--Olof

dcornejo:
I'm looking for end-to-end protection - in our code the XML can bounce around and get URI encoded several times, but when it gets to the destination we don't know how many times it's been encoded and therefore we can't be certain what the original was.

For example if the user provides a description (very contrived) of "&lt;something>" it would get reencoded as "&amp;lt;something&gt;" and we could decode it until there are no more decodings possible, but in this case we'd get the wrong answer of "<something>"

we know what the endpoints of the message are, so we take the responsibility to add/remove the CDATA rather than try to leave it to clixon to guess at.

olofhagsand:
Yes, but in that case you need to add code in the XML encoding (and decoding) so that the CDATA expression is skipped in those cases. Otherwise, when you save to file, for example, the payload will be encoded also inside the CDATA.
Example - CDATA text contains a single '&':
<rpc><edit-config><target><candidate/></target><config><text><![CDATA[&]]></text></config></edit-config></rpc>]]>]]>
Save to file:
<config>
   <text>&lt; ![CDATA[&amp; ]]&gt; </text>
</config>
But I can pull it into develop branch, and then fix it there. Any specific reason you wanted it in master?

dcornejo:
sorry, intended for develop branch.

thanks!

olofhagsand:
Merged into develop branch

PR #32, opened 2018-07-03
make specification of an operation compliant with RFC8040 section 3.6

### LINKED ISSUE #31: RPC operation should use module name rather than prefix
see @dmittryvvv comment on closed https://github.com/clicon/clixon/issues/30

RFC-8040 states the proper specification for an operation is

`POST /restconf/operations/<module_name>:<rpc_procedure> HTTP/1.1
`

but clixon requires

`POST /restconf/operations/<module_prefix>:<rpc_procedure> HTTP/1.1
`

dcornejo:
RFC 8030 section 3.6 indicates that an operation is identified by both module name and operation ID. The latest code used the prefix rather than the module name, which I believe is contrary to the RFC. This pull request removes the (possibly) erroneous check of the prefix and changes the code to require a module name match to resolve #31

### DIFF
--- a/lib/src/clixon_yang.c
+++ b/lib/src/clixon_yang.c
@@ -2137,16 +2137,12 @@ yang_abs_schema_nodeid(yang_spec    *yspec,
     }
     if (ymod == NULL){ /* Try with topnode */
 	if ((ys = yang_find_topnode(yspec, id, YC_SCHEMANODE)) == NULL){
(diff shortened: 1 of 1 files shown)


olofhagsand:
I needed to make a larger and more specific change for restconf operations GET and PUT, see patch above. 
Thanks for detecting this major divergence from RFC 8040!

PR #45, opened 2018-10-24
Minor changes on NETCONF implementation

SCadilhac:
* No space after ampersand escaped characters
* Add writable-running capability
* Announce base:1.0 capability (because we support only legacy framing system)
* No quotes around revision value

### DIFF
--- a/apps/netconf/netconf_hello.c
+++ b/apps/netconf/netconf_hello.c
@@ -163,8 +163,8 @@ netconf_create_hello(clicon_handle h,
     add_preamble(cb);
     cprintf(cb, "<hello xmlns=\"urn:ietf:params:xml:ns:netconf:base:1.0\">");
     cprintf(cb, "<capabilities>");
-        cprintf(cb, "<capability>urn:ietf:params:netconf:base:1.1</capability>");
-    if (xml_chardata_encode(&encstr, "urn:ietf:params:netconf:capability:yang-library:1.0?revision=\"%s\"&module-set-id=%s",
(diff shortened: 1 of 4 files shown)


olofhagsand:
@SCadilhac thanks!!

I have cherry-picked and pushed into develop:
- Add conformance-type to modules-state (Had to enhance test/test_feature.sh)
- Allow new lines in prompts

But I have some comments on the others:
(1) No space after ampersand escaped characters - 
I strictly followed https://www.w3.org/TR/2008/REC-xml-20081126/#syntax where it says eg: 
       single-quote character (') may be represented as " &apos; " 
Note the space after the semi-colon. I am happy to remove the space, seems strange to me, but it is there in the W3C spec. Any comment?
2) Add writable-running capability. Announce base:1.0 capability
writable-running: OK, strictly speaking it is supported in the sense that you can write to the running config datastore. But there is no commit action associated if you write directly to running, ie no backend actions are triggered as when you write to candidate and then do commit. I don't see any discussion on that in RFC6241. I.e., what is the semantics of writable-running? But I would expect that you would want backend plugins to be triggered and they currently are not. So I question whether the "writable-running" capability is actually supported?
base:1.0 OK,
quotes around version number: OK
3) Add xmlns to the element in xml_merge
I agree to the error, but have some problems with the patch due to (my) style of programming using labels, I use ok: only as a positive exit from the function with no (or only minor) side-effects.
As this seems to be a global error,I think I may need to look at this generally.

PR #55, opened 2018-11-06
changes for FreeBSD compatibility

dcornejo:
ETIME shows in many distributions as "timer expired", I think ETIMEDOUT is more appropriate as it means "connection timed out" which is more correct for the intent of this error. Plus it allows compilation on FreeBSD. ETIMEDOUT is in both CentOS and Ubuntu.

add sys/time.h to clixon_handle.c. This is required to compile on FreeBSD, the header file exists in both Ubuntu and CentOS and so this change compiles in all three.

### DIFF
--- a/apps/netconf/netconf_main.c
+++ b/apps/netconf/netconf_main.c
@@ -294,7 +294,7 @@ static int
 timeout_fn(int s,
 	   void *arg)
 {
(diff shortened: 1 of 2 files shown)


dcornejo:
replaced by pull request #76

PR #76, opened 2019-02-23
FreeBSD compatibility & pedanticism patches 

dcornejo:
this pull request supersedes pull request #55 

use ETIMEDOUT rather than ETIME - this seems more the intent of its
usage here and it compiles under FreeBSD & Linux.
in yang_spec_parse_file() the function basename() is not declared,
the declaration is in libgen.h, so include it. also basename() takes
char * for its argument not const char *, yang_spec_parse_file()
has to be changed to char *filename to compile w/o warnings

these changes have been checked under both Linux and FreeBSD, merging them would simplify the FreeBSD port and remove warnings from both FreeBSD and Linux

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.o
 *.so.*
+*.so
 *_parse.tab.c
 *_parse.tab.h
 lex.*_parse.c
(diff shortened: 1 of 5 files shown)


olofhagsand:
Thanks. I will extend the CI to freebsd to avoid this.

PR #85, opened 2019-06-14
Startup extraxml file load fix

008agent:
When backend is started, startup_extraxml is called without any
checks. It causes "startup" config being overrided(not merged)
with "tmp" which is null due to nullprt file


olofhagsand:
The failure case is valid, but required some more work, in particular, the extra-xml function also triggers  the reset callbacks. Instead inserted a check if the tmp db is empty before a validation is made.
See:
https://github.com/clicon/clixon/commit/3bad8bc874fef8e885db19147db3f77699fb46ab

PR #89, opened 2019-06-19
api_path2xml_vec strict mode check on list key length mismatch

008agent:


olofhagsand:
Cornercase where a restconf api-path should not enforce all key elements.

PR #94, opened 2019-10-30
Add missing includes

rbgarga:
Added missing includes on clixon_string.h:

* stdlib.h for malloc()
* string.h for strncpy()

### DIFF
--- a/lib/clixon/clixon_string.h
+++ b/lib/clixon/clixon_string.h
@@ -36,6 +36,9 @@
 #ifndef _CLIXON_STRING_H_
 #define _CLIXON_STRING_H_
 
+#include <stdlib.h>
+#include <string.h>
+
 /* Struct used to map between int and strings. Typically used to map between
  * values and their names. Note NULL terminated
  * Example:


olofhagsand:
@rbgarga I try to not to have includes in include-files except for the top-level include, eg clixon.h(.in). The reason being not totally consistent but involves trying to keep the C include structure flat. Would it work adding them in clixon.h(.in) instead?

rbgarga:
I believe it will be enough. Thanks!

olofhagsand:
@rbgarga @dcornejo I am trying to understand the original problem (so I can write a regression test). What is the original problem, what is solved by adding the include files and on what platform?

olofhagsand:
Added fix in branch: release-4.2. If you can verify I will stamp a tag.

dcornejo:
On CentOS/gcc: The exact problem is that malloc() and strncpy() were not prototyped in the included system headers.

We actually hit this problem in 2 places, one was in the 4.2 merge with malloc() which I fixed by including stdlib.h in my code. The other place was with clixon 4.0 (approximately) in moving from CentOS 7 to 8 that was fixed in this pull request in clixon code.

I don't have the code in front of me to look, but I believe that a combination of changes to clixon and our environment where the system headers changed that exposed this.

olofhagsand:
OK, gcc on CentOS 8, thanks. But it cant be clixon_string.c, because it already has stdlib.h and string.h included? 
It must be some other .c file? This is what puzzles me.

rbgarga:
> OK, gcc on CentOS 8, thanks. But it cant be clixon_string.c, because it already has stdlib.h and string.h included?
> It must be some other .c file? This is what puzzles me.

We have .c files that includes <clixon_string.h>.  When building gcc complains it's using malloc or snprintf without a definition.  Then we need to include <string.h> or <stdlib.h> to our .c files even if they are not consuming it.

olofhagsand:
Aah now I go it: there is an inline statement in the .h: `static inline char * strdup4(char *str)` that calls malloc and string funtions inline, which explains it.
Got it, thanks for detecting it!

PR #103, opened 2020-01-30
fix problem validating leafref in state data referencing config data

dcornejo:
when the config data was merged in from_client_get() the xpath
excluded the config data, preventing the leafref from being validated.
change is to not filter the config data out at all.


olofhagsand:
Fixed by this: https://github.com/clicon/clixon/commit/a58f1872ad26f4521b118000a2de545d019ec6d1

PR #114, opened 2020-06-24
Changed code so that it will be compatible with c++

shmuelnatan:
(using g++ insted of gcc). Currently the examples are compiling properly.


shmuelnatan:
The things I changed in general:
1. Variables that are key words in c++ were changed.
2. Implicit conversions were changed to explicit conversions.
3. Added #ifdef extern "C" to public header files.
4. constant strings that were in a char* variable were changed to const char*.

example/main/example_backend.c
```diff
@@ -565,7 +565,7 @@ example_upgrade(clicon_handle    h,
 	    goto done;
 	/* Remove them */
 	/* Loop through all nodes matching mypath and change theoir namespace */
-	for (i=0; i<xlen; i++){
+	for (i=0; i<(int)xlen; i++){
```
dima1308, line 568:
```suggestion
	for (size_t=0; i<xlen; i++){
```

dima1308: I suggest to do it for all similar places

example/main/example_backend.c
```diff
@@ -269,18 +269,18 @@ example_rpc(clicon_handle h,            /* Clicon handle */
 {
     int    retval = -1;
     cxobj *x = NULL;
-    char  *namespace;
+    char  *_namespace;
```
dima1308, line 272:
```suggestion
    char  *namespace_;
```

olofhagsand:
One comment and one question:
1) Can you please not prepend a variable parameter with "_" eg when renaming namespace -> _namespace, instead renaming to "ns" or something more terse.  Same with "explicit" -> "_explicit", rather use eg "expl". The reason is we use prepended "_" as private/local variables, and these are public variables.
2) Question: I understand why parameter renaming and adding "const" qualifiers in .h files may spread into the .c files, but some changes in the .c files seem to indicate you want to compile them with c++? Such as the many examples in example_backend.c/example_backend_nacm.c. Another example is clixon_json.c. Why would you want to do that, is it not enough to make the signature changes in the .h:s, and then eg make "const" changes in the .c files only?
I know I proposed to make a C++ example, but I would rather you make a separate (very simple) C++ example in that case.

shmuelnatan:
Hi,
So I changed the names I gave to variables from having a _ symbol, to an abbreviation (like ns for namespace) or a shorter version of the word (like expl for explicit and nw for new).
I also deleted all the examples (so that they'll be unchanged), besides example_backend.c that was changed to support c++.

example/main/example_backend.c
```diff
@@ -269,18 +269,18 @@ example_rpc(clicon_handle h,            /* Clicon handle */
 {
```
shmuelhazan, line 269: This file should have been left untouched.

example/main/example_restconf.c
```diff
@@ -67,8 +67,8 @@ static int basic_auth = 0;
  @note what is copyright of this?
  */
```
shmuelhazan, line 68: Same for all files under example main:
```
git checkout origin/master -- example 
```

apps/cli/clixon_cli.h
```diff
@@ -60,3 +63,7 @@ int plugin_start(clicon_handle h, int argc, char **argv);
 int plugin_exit(clicon_handle h);
 
 #endif  /* _CLIXON_CLI_H_ */
+
+#ifdef __cplusplus
+} /* extern "C" */
+#endif
```
dima1308, line 69: for all files, better to have a new line at the end...
it may cause compilation warnings

PR #118, opened 2020-07-01
Changed code so that it will be compatible with c++.

shmuelnatan:
1. Added #ifdef extern "C" to public header files.
2. Changed variables that are key words in c++, to different names (like ns instead of namespace).
3. Changed char* to const char* for constant strings.


shmuelnatan:
Updating the code to be compatible with c++
1. Added #ifdef extern "C" to public header files.
2. Changed variables that are key words in c++, to different names (like ns instead of namespace).
3. Changed char* to const char* for constant strings.
4. Added a test case named test_c++.sh that compiles a code with g++.

olofhagsand:
See comments on slack

PR #123, opened 2020-07-28
Don't call upgrade callbacks for module with no revision defined

008agent:


olofhagsand:
lixon_yang_module.c: In function 'mod_ns_upgrade':

clixon_yang_module.c:414:2: warning: this 'if' clause does not guard... [-Wmisleading-indentation]

  414 |  if ((yrev = yang_find(ymod, Y_REVISION, NULL)) == NULL)

      |  ^~

clixon_yang_module.c:416:6: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'

  416 |      goto done;

You may have forgotten {} ?

PR #125, opened 2020-08-05
sort the XML before dumping when using -q switch

dcornejo:
when using the -q switch of clixon_backend (upgrade & dump) - sort the XML before dumping it to guarantee a consistent order in the output.


olofhagsand:
Breaks the upgrade-quit test, but I can fix that after merge.

PR #132, opened 2020-08-31
fix show command.

shmuelnatan:
1. all list is displayed with show <list_name>.

2. leaf is displayed by doing show <leaf_name> (without adding the contents of the leaf).


shmuelnatan:
This PR doesn't completely solve the problem. I will open a new PR that does solve the problem.

PR #135, opened 2020-09-07
Fix show and set command 

shmuelnatan:
1. All list is displayed with show <list_name>.
2. Leaf is displayed by doing show <leaf_name> (without adding the contents of the leaf).
3. Created three generated trees @datamodel (for set command) @datamodelshow (for show configuration command), @datamodelshowstate (for show state command).

### DIFF
--- a/apps/cli/cli_generate.c
+++ b/apps/cli/cli_generate.c
@@ -776,6 +776,7 @@ yang2cli_list(clicon_handle      h,
     int           retval = -1;
     char         *helptext = NULL;
     char         *s;
+	int 		 list_has_callback;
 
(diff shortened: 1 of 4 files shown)


olofhagsand (commented):
I think the approach is OK.

Bit example and some tests break. 
Example should use @datamodelshow any place where auto_cli_show() is called in the clispec.

A test failure: test_feature.sh:
yang2cli:72: Error: syntax error: at or before: '}'
which when inspected is because there is an extra }, here:
```
           description("Textual description of the RIB.") ,overwrite_me("/ietf-routing:routing/ribs/rib=%s/description");
         }
         }
      }
  } <---
```

PR #137, opened 2020-09-10
Fix show and set command 

shmuelnatan:
1. All list is displayed with show <list_name>.
2. Leaf is displayed by doing show <leaf_name> (without adding the contents of the leaf).
3. Created three generated trees @datamodel (for set command) @datamodelshow (for show configuration command), @datamodelstate (for show state command).

### DIFF
--- a/apps/cli/cli_generate.c
+++ b/apps/cli/cli_generate.c
@@ -776,6 +776,7 @@ yang2cli_list(clicon_handle      h,
     int           retval = -1;
     char         *helptext = NULL;
     char         *s;
+	int 		 list_has_callback;
 
     cprintf(cb, "%*s%s", level*3, "", yang_argument_get(ys));
     if ((yd = yang_find(ys, Y_DESCRIPTION, NULL)) != NULL){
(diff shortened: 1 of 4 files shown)


olofhagsand:
While query (?) seems to work, there are cases with lists that do not work. Seems to work for "text" but not for "xml" or "json". 
Example:
```
olof@alarik> show configuration text table parameter # THIS WORKS
parameter {
    name a;
    value 99;
}
parameter {
    name b;
    value 99;
}
olof@alarik> show configuration xml table parameter # THIS DOES NOT WORK
<parameter>
   <name>a</name>
   <value>99</value>
</parameter>
```
Maybe this is due to some other error?
I am prepared to pull this OK since it passes all regression tests, if you can continue looking at these other cases?

shmuelnatan:
I added a fix for JSON and XML. It works. You need to merge the `apps/cli/cli_show.c ` file in b02f8afa9ee5435ac7f9c2648f96c99738201abd commit (I don't have write access so I can't do it).

olofhagsand:
fixed conflicts and pulled.

PR #149, opened 2020-11-07
Topic benavrhm rfc8527a 20201103

benavrhm:


apps/restconf/restconf_root.c
```diff
@@ -441,7 +490,41 @@ api_root_restconf(clicon_handle h,
...
+	 * running/cadidate is hardcoded into the clixon implementation. 20201104 */
+	ietf_ds_t ds = IETF_DS_NONE;
+
+	if (4 > pn) { /* Malformed request, no "ietf-datastores:<datastore>" component */
+           restconf_notfound(h, req);
+	   goto done;
+        }
+
+	/* Assign ds; See https://tools.ietf.org/html/rfc8342#section-7 */
+	if      (0 == strcmp(pvec[3], "ietf-datastores:running"))
```
olofhagsand, line 508: please remove spaces after "if"

apps/restconf/restconf_root.c
```diff
@@ -441,7 +490,41 @@ api_root_restconf(clicon_handle h,
     }
     else if (strcmp(api_resource, "data") == 0){ /* restconf, skip /api/data */
 	if (api_data(h, req, path, pcvec, 2, qvec, indata,
-		     pretty, media_out) < 0)
+		     pretty, media_out, IETF_DS_NONE) < 0)
+	    goto done;
+    }
+    else if (strcmp(api_resource, "ds") == 0) {
```
olofhagsand, line 496: I have a basic/stupid? question on how the "ds" resource is specified and accessed in the web server, eg nginx or evhtp code.
Why is this a separate data path than the "data" resource?

olofhagsand (commented):
Thanks for reviewing the tests.

olofhagsand:
There are some conflicts due to some merging I did on master,  resolve them by choosing the master variant.

test/test_perf_state.sh
```diff
@@ -139,7 +139,9 @@ expecteof "time -p $clixon_netconf -qf $cfg" 0 "<rpc $DEFAULTNS><commit/></rpc>]
 new "netconf get test single req"
```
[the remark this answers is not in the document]
benavrhm: Hmm. Seems to be a mistake in how I ordered the patches (the commits). After the next commit that adds the new ietf yang files this should work. I just now retested.

apps/restconf/restconf_root.c
benavrhm: Will do.

apps/restconf/restconf_root.c
benavrhm: This is the "new way" of specifying datastores in RESTconf as per https://tools.ietf.org/html/rfc8527 and https://tools.ietf.org/html/rfc8342. AFAIK the reason that no web server change is required is because the entire request is encoded in XML as an rpc and passed transparently to the underlying netconf layer.

apps/restconf/restconf_root.c
benavrhm: Pushed fix now.

benavrhm:
AFAIK I have completed handling all of the comments and the ball is in your court. Thanks.

apps/restconf/restconf_root.c
olofhagsand: Yes, I missed that clause in RFC8572, my mistake

PR #151, opened 2020-11-13
make xml2xpath1() less picky

dcornejo:
this function would never traverse up an XML tree from a node that
was not bound to a YANG model. this prevents the function from
working in some cases where it could.

### DIFF
--- a/lib/src/clixon_xml_map.c
+++ b/lib/src/clixon_xml_map.c
@@ -1503,8 +1503,7 @@ xml2xpath1(cxobj *x,
     char         *b;
     enum rfc_6020 keyword;
     
-    if ((xp = xml_parent(x)) != NULL &&
-	xml_spec(xp) != NULL)
+    if (xp = xml_parent(x))
 	xml2xpath1(xp, cb);
     /* XXX: sometimes there should be a /, sometimes not */
     cprintf(cb, "/%s", xml_name(x));


olofhagsand:
There is a non-fatal side-effect seen in the tests, it seems the patch included the root-element "/config" in the xml2xpath which should not be there. So it is not enough to just check for that parent exists.
Instead either the root element should be excluded (ie xml_parent() is NULL), or (worse) check that its name is "config"
From: https://travis-ci.org/github/clicon/clixon/builds/743452585:
```
Expected: ^<rpc-reply xmlns="urn:ietf:params:xml:ns:netconf:base:1.0"><rpc-error><error-type>protocol</error-type><error-tag>operation-failed</error-tag><error-app-tag>too-few-elements</error-app-tag><error-severity>error</error-severity><error-path>/c/a1</error-path></rpc-error></rpc-reply>]]>]]>$
Received: <rpc-reply xmlns="urn:ietf:params:xml:ns:netconf:base:1.0"><rpc-error><error-type>protocol</error-type><error-tag>operation-failed</error-tag><error-app-tag>too-few-elements</error-app-tag><error-severity>error</error-severity><error-path>/config/c/a1</error-path></rpc-error></rpc-reply>]]>]]>
```

olofhagsand:
I needed to slightly patch it some more taking the "grandparent" into account.

PR #165, opened 2020-12-28
datastore_read: bind yang when reading from cache

baruchsiach:
The cached xml version might be missing the yang information. This
breaks netconf validation of the running database:

yangcli admin@192.168.0.1> validate source=startup

RPC Error Reply 3 for session 3:

rpc-reply {
  rpc-error {
    error-type application
    error-tag operation-failed
    error-severity error
    error-message 'No spec found radio-dn'
  }
}

Make sure that the cached database respects the yb parameter.

### DIFF
--- a/lib/src/clixon_datastore_read.c
+++ b/lib/src/clixon_datastore_read.c
@@ -675,6 +675,10 @@ xmldb_get_cache(clicon_handle    h,
     else
 	x0t = de->de_xml;
 
+    if (yb == YB_MODULE && !xml_spec(x0t))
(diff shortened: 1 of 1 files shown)


olofhagsand:
Can you please describe a use-case when this happens?
From the example above it seems to be when you access startup (using NMDA?)
One reason is one need to handle the case of retval = 0 of xml_bind_yang() as well as make a proper regression test.

baruchsiach:
The yangcli command in the commit log shows the error. The setup that produces this error is otherwise functional.
It seems inconsistent that the yb parameter is not taken into account when xml is read from cache.

PR #167, opened 2021-01-06
completing the function of hide for autocli-op in yang files 

shmuelnatan:
1. Added the hide functionality to containers (until now it was only implemented for lists and leaves).

2. Added the functionality to showing the database (configuration and state), meaning, if there's a hide for a parent in a yang module, then you won't see it and its "offspring", no matter if it's state data or configuration.


shmuelnatan:
@olofhagsand This PR failed some tests. I think it's because I changed the functionality of hide for autocli-op in yang files.

PR #169, opened 2021-01-10
completing the function of hide for autocli-op in the yang files

shmuelnatan:
Changed the "hide" command to "hide-auto-completion", because it's hidden from the auto-completion but not from the database (useful for data that we don't want to be changed by user. for example, SW version).

Added a "hide-database" command in order to hide it from the database but not from the auto-completion (useful for setting passwords and not exposing them).

Added a "hide-database-auto-completion" command in order to hide it from the database and from the auto-completion (useful for data that we don't want to expose to user).

Added the functionality to containers too (until now it was only implemented for lists and leaves).

clicon/cligen#61 is the PR in Cligen that completes this PR


olofhagsand:
(From slack)
OK now I got it, thanks!
- I have a problem with renaming the "hide" keyword because it breaks backward compatibility.
- Regarding that concatenation does not work - I can have a look at that, I would rather we fix the original problem. There are potential other flags and it would be much more convenient to concatenate them.
- I see the need for hide-database, thanks. The only problem is that this is a clixon not a cligen feature (cligen has no concept of "database"). I dont know, maybe "hide-show" is a better term?
- Also I would also want to tie this to YANG as described here: https://clixon-docs.readthedocs.io/en/latest/cli.html#hidden-commands, that is, if "hide-database" is added it could be good to add it as an extension to YANG as well, so that you can mark in YANG which commands you want hidden in the auto-cli.
I could help with some of the bullets (such as the last), but would like to have discussion on the comments.

PR #206, opened 2021-04-19
completing the function of hide for autocli-op in the yang files

shmuelnatan:
1. Added a "hide-database" command in order to hide it from the database but not from the auto-completion (useful for setting passwords and not exposing them). 
2. Added a "hide-database-auto-completion" command in order to hide it from the database and from the auto-completion (useful for data that we don't want to expose to user).
3. Kept backwards compatibility with not changing the wording for the "hide" command (which basically hides the auto completion of option of commands).

The following PR https://github.com/clicon/cligen/pull/62 completes this PR.

### DIFF
--- a/apps/cli/cli_auto.c
+++ b/apps/cli/cli_auto.c
@@ -120,6 +120,28 @@ cvec_append(cvec *cvv0,
     return cvv2;
 }
 
+/*! x is element and has exactly one child which in turn has none 
+ * @see child_type in clixon_json.c
+ */
+static int
+tleaf(cxobj *x)
+{
+    cxobj *xc;
+
+    if (xml_type(x) != CX_ELMNT)
+	return 0;
+    if (xml_child_nr_notype(x, CX_ATTR) != 1)
+	return 0;
+    /* From here exactly one noattr child, get it */
+    xc = NULL;
+    while ((xc = xml_child_each(x, xc, -1)) != NULL)
+	if (xml_type(xc) != CX_ATTR)
+	    break;
+    if (xc == NULL)
+	return -1; /* n/a */
+    return (xml_child_nr_notype(xc, CX_ATTR) == 0);
+}
+
 /*! Print an XML tree structure to an output stream and encode chars "<>&"
  *
  * @param[in]   xn          clicon xml tree
--- a/apps/cli/cli_auto.c
+++ b/apps/cli/cli_auto.c
@@ -521,10 +521,10 @@ cli_auto_show(clicon_handle h,
 	switch (format){
 	case FORMAT_XML:
 	    if (isroot)
-		stdout(xp, 0, pretty, 1);
+		xml2file(xp, 0, pretty, fprintf);
 	    else{
 		while ((xc = xml_child_each(xp, xc, CX_ELMNT)) != NULL)
-		    stdout(xc, 0, pretty, 1);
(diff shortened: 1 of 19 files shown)


olofhagsand:
I have two comments. One is simple, and the other is difficult.

1) Can you please document the extensions hide-database / hide-database-auto-completion?
One way to do that is in clixon-lib@2021-03-08.yang (in the description of the autocli-op extension where "hide" is already described).

2) Also, I am somewhat reluctant to add the following conditional:
```
	/* Look for autocli-op defined in clixon-lib.yang */
	if (yang_extension_value(xml_spec(x), "autocli-op", CLIXON_LIB_NS, &opext) < 0) {
		goto ok;
	}
	if ((used_for_show_command == 1) && (opext != NULL) && ((strcmp(opext, "hide-database") == 0) || (strcmp(opext, "hide-database-auto-completion") == 0))){
		goto ok;
}
```
in the "Low-level" function xml2file_recurse(). The reason is that those extensions are more high-level and it would be better to make that masking in a higher level callback, such as an auto-cli "aware" function. 

However, I am not sure how to that, I was thinking that one could do an `autocli_output` function but that does not work since it does not take any arguments apart from printf:s.

Do you know exactly which calls to clicon_xml2file() are needed to intercept for this filtering? Is it all calls or just a specific subset?

Let us have a discussion about this, maybe on slack. One may need to add some abstraction here, like a new callback.

shmuelnatan:
I finished your first request.
I don't understand what you meant here 
`Do you know exactly which calls to clicon_xml2file() are needed to intercept for this filtering? Is it all calls or just a specific subset?`
`yang_extension_value(...)` and `xml_spec(...)` are the only functions needed for the interception.

shmuelnatan:
I don't know if my solution is what you intended, but I moved the examination of the yang extension from clixon_xml_io.c to auto_cli.c 
I checked the code and it works.

apps/cli/cli_auto.c
```diff
@@ -120,6 +120,114 @@ cvec_append(cvec *cvv0,
...
+ *
+ * @param[in]   xn          clicon xml tree
+ * @param[in]   level       how many spaces to insert before each line
+ * @param[in]   prettyprint insert \n and spaces tomake the xml more readable.
+ * @param[in]   fn          Callback to make print function
+ * @see clicon_xml2cbuf
+ * One can use clicon_xml2cbuf to get common code, but using fprintf is
+ * much faster than using cbuf and then printing that,...
+ */
+int xml2file (cxobj *xn, int level, int prettyprint, clicon_output_cb *fn)
```
olofhagsand, line 133: Please make it static and add something like "autocli_" prefix, eg: static int autocli_xml2file

olofhagsand:
Yes, that is better. Can you also please do the same thing with xml2cli_recurse() in clixon_xml_map.c, that is, add a specialized function in cli_auto.c?
Then I will merge it, and I will think about how to make that more generic by some callback or similar.

apps/cli/cli_auto.c
shmuelnatan: This function is also used in show_cli.c so it can't be static. Do you want to make it an extern function?

olofhagsand: Oh, did not see that. Fine just please add prefix. No extern please.

PR #217, opened 2021-05-02
Netconf Chunks (RFC6242 - Section 4.2)

s-bauer:
Hey Olof, another PR from my end. It's not yet done and needs some more work to correctly detect whether the chunking format or the traditional end-of-message-based mechanism is used. I just wanted you to have an look at it to maybe get some early feedback.


s-bauer:
@mager-m FYI

s-bauer:
Open Questions:
- Should we strictly use the traditional approach for NETCONF 1.0 and the chunk approach for NETCONF 1.1 like the official standard says or should we allow both formats (like I'm currently doing it)?
- How should the response encoding be? Depending on the request format? Like if the client is sending the request as chunked, then the response should be chunked as well?

Open Topics:
- Chunk Validating (size, ...)

olofhagsand:
It needs to be backward compatible in some way.
Section 4.3 says is if `:base.1.1` capability is not advertised by both peers the old framing should be used.
But clixon advertises `:base.1.1` now. 
My proposal would be to add a config option which is by default false and follows the RFC strictly, but can be set to true to be backward compatible and use the old framing `]]>]]>` even if `:base.1.1` is advertised. 
In this way, users can set this config variable and work with moving over to chunked and the reset it.

apps/netconf/netconf_main.c
```diff
@@ -523,10 +581,24 @@ netconf_input_cb(int   s,
 		}
 		cbuf_reset(cb);
 	    }
+
+            if(chunkRemainingLength == -1) {
+                if(detect_netconf_chunk_start(buf[i], &chunkStartState, &chunkLength)) {
+                    clicon_log(LOG_INFO, "%s: detected chunk with size %i", __FUNCTION__, chunkLength);
```
olofhagsand: LOG_INFO syslog may be too chatty in a live system?
Can you use clicon_debug(1,..) ?

apps/netconf/netconf_main.c
```diff
@@ -449,12 +449,16 @@ netconf_input_frame(clicon_handle h,
     return retval;
 }
 
-/*! Look for an endtag in a cubf
- *  @param[in]     cb           The cbuf holding the received message
- *  @retval        0            No end tag has been detected
- *  @retval        1            End tag has been detected
+/*! @brief          Detect a given character sequence at the end of a \b cbuf
```
olofhagsand, line 452: brief/detail doxygen not used anywhere else in clixon.
doxygen seems to figure that out itself giving the first line as brief and after a \n uses the remaining text as detail?

olofhagsand (commented):
Also before you are done, please also run the memory leak check, using test/mem.sh netconf on the test suite.

s-bauer:
Hey @olofhagsand, sorry for the delay! I've had the time to further work on this and made the following changes:
- Refactored Makefile for netconf
- Added a capability context that keeps track of the announced server and client capabilitites. (Server capabilities are not yet used)
- Adjusted `netconf_output_encap` so that it uses the chunked framing mechanism when the client announced the netconf 1.1 capability.
-Added a `CLICON_NETCONF_FRAMING_FORCEENDOFMESSAGE` that can be used to suppress the new chunked framing mechanism (not yet implemented)

Open Topics:
- Tests (including memory leak test)
- Using server capabilitites of capability context
- Use the `CLICON_NETCONF_FRAMING_FORCEENDOFMESSAGE` flag
- Capability Context should support capability parameters (e.g. `urn:ietf:params:netconf:capability:yang-library:1.0?revision=...`)

olofhagsand:
Memory leaks should be covered by the test/mem.sh script using valgrind for netconf/restconf/backend.

olofhagsand:
Chunked framing is now included in the upcoming 5.7 release, closing this PR

PR #239, opened 2021-06-15
Fixed memleak when building argv for restconf

nowaits:


olofhagsand:
Thanks, but fixed this as part of the release 5.2 memory tests.

PR #243, opened 2021-07-06
add xpath prefix for xml2xpath if available

nowaits:


olofhagsand:
While I agree on the error statement, the patch in `xml2xpath1()` only solves it partly. For example, in the return in text_minmax.sh:

```
<rpc-reply xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" message-id="42"><rpc-error><error-type>protocol</error-type><error-tag>operation-failed</error-tag><error-app-tag>too-few-elements</error-app-tag><error-severity>error</error-severity><error-path>/ex:c/a1</error-path></rpc-error></rpc-reply>]]>]]>
```

there are two issues:
1) the prefix "ex" needs to be declared in the XML for it to be well-formed, such as with a `xmlns:ex:"..."` statement.
2) the xpath prefix needs to be used in all symbols in the xpath, not just the uppermost, eg it should be `/ex:c/ex:a1`
Therefore, without looking at how to fix it, the message should be something like:
```
<rpc-reply xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" message-id="42"><rpc-error><error-type>protocol</error-type><error-tag>operation-failed</error-tag><error-app-tag>too-few-elements</error-app-tag><error-severity>error</error-severity><error-path xmlns:ex="urn:example:clixon">/ex:c/ex:a1</error-path></rpc-error></rpc-reply>]]>]]>
```
Likewise in test_yang_deviation.sh

I think problem nr 2 can be fixed by 
1) (best) finding the _namespace_ (not prefix) and checking if that is declared in the xml node at hand using `xml2prefix()`, if yes then use the retirned prefix, else create new.
2) if you want to stick with the prefix, then check with `xml2ns()` if it exists and bound to the right namespace, and create it if it is not present.

PR #248, opened 2021-07-25
split up long function, use safe string handling functions

alanyanigersiklu:
For review by Shmuel


apps/restconf/restconf_methods.c
```diff
@@ -581,6 +581,367 @@ api_data_write(clicon_handle h,
 } /* api_data_write */
 
 #ifdef YANG_PATCH
+
+char * init_str()
```
shmuelhazan, line 585: if you need safe, resizeable vector, you probably can use cligen's cbuf

apps/restconf/restconf_methods.c
```diff
@@ -581,6 +581,367 @@ api_data_write(clicon_handle h,
...
+    }
+    for (int j = 0; j < veclen; j++) {
+        cxobj *xn = vec[j];
+        tmp_val = xml_body(xn);
+    }
+    cpy_str(val, tmp_val, TEMP_STR_MALLOC_SIZE);
+    return 0;
+}
+
+int do_replace (
```
shmuelhazan, line 660: We can have another function called do_replace. I suggest to either make it static, or rename it.

apps/restconf/restconf_methods.c
```diff
@@ -581,6 +581,367 @@ api_data_write(clicon_handle h,
...
+        clicon_handle h,
+	    void         *req,
+	    int           pi,
+	    cvec         *qvec,
+	    int           pretty,
+	    restconf_media media_out,
+	    ietf_ds_t     ds,
```
shmuelhazan, line 667: unclear what are the arguments. Can you please add docs?

apps/restconf/restconf_methods.c
```diff
@@ -581,6 +581,367 @@ api_data_write(clicon_handle h,
...
+          if (value_vec[k] != NULL) {
+              value_vec_tmp = xml_dup(value_vec[k]);
+              xml_addsub(x_simple_patch, value_vec_tmp);
+          }
+      }
+      cbuf* cb = cbuf_new();
+      xml2json_cbuf(cb, x_simple_patch, 1);
+
+      // Some ugly text processing to get the JSON to match what api_data_post() expects
+      char *json_simple_patch_tmp = cbuf_get(cb);
```
shmuelhazan, line 708: Can we modify api_data_post to support this syntax? 
In any case, I would suggest adding more explicit description to hacks.

apps/restconf/restconf_methods.c
```diff
@@ -581,6 +581,367 @@ api_data_write(clicon_handle h,
...
+      for (int l = strlen(target_val); l>= 0; l--) {
+          if (target_val[l] == '/') {
+              idx = l;
+              break;
+          }
+      }
+      cpy_str(post_req_uri, target_val, idx);
+      cat_str(simple_patch_request_uri, post_req_uri, TEMP_STR_MALLOC_SIZE);
+      free_mem((void *)post_req_uri);
+      for (int l = strlen(json_simple_patch); l>= 0; l--) {
```
shmuelhazan, line 736: unclear what is this for?

apps/restconf/restconf_methods.c
```diff
@@ -581,6 +581,367 @@ api_data_write(clicon_handle h,
...
+              brace_count++;
+              if (brace_count == 2) {
+                  json_simple_patch[strlen(json_simple_patch)] = '[';
+              }
+          }
+          json_simple_patch[strlen(json_simple_patch)] = c;
+      }
+
+      /* strip /... from end */
+      char *post_req_uri = init_str();
```
shmuelhazan, line 722: can we move this outside as a function? (e.g. cbuf_strip_slash)

apps/restconf/restconf_methods.c
```diff
@@ -581,6 +581,367 @@ api_data_write(clicon_handle h,
...
+	    ietf_ds_t     ds,
+        char* simple_patch_request_uri,
+        int value_vec_len,
+        cxobj** value_vec,
+        cxobj * value_vec_tmp,
+        cxobj *x_simple_patch,
+        char *patch_header
+      )
+{
+     for (int k = 0; k < value_vec_len; k++) {
```
shmuelhazan, line 770: can you please add documentation?

shmuelhazan (changes requested):

olofhagsand:
Please use the merged version in master as your base, eg that uses `#define YANG_PATCH` to enable patch in include/clixon_custom.h rather than in apps/restconf/restconf_api.h. 
For example, use `git rebase master` or something equivalent.
Also (as pointed out by @shmuelhazan ^) please do not construct new string functions within restconf_methods.c. Either use existing string functions, eg the cligen_buf:s, or extend the existing string functions such as in clixon_string.[ch]. 
The reason is code divergence.
My primary concern with this code currently is lack of regression tests (as placed under test/). When such a test is in place, it can be used for memory tests (see test/mem.sh), cross-platform tests (test/cicd) and fuzzing (fuzz/restconf).

apps/restconf/restconf_methods.c
```diff
@@ -581,6 +583,409 @@ api_data_write(clicon_handle h,
...
+      cbuf* val_tmp = cbuf_new();
+      cbuf_append_str(val_tmp, cbuf_get(val));
+      int idx = cbuf_len(val_tmp);
+      for (int l = cbuf_len(val_tmp); l>= 0; l--) {
+          if (cbuf_get(val_tmp)[l] == '/') {
+              idx = l;
+              break;
+          }
+      }
+      cbuf* val_tmp_2 = cbuf_trunc(val_tmp, idx + 1);
```
vsharapov-udelphi, line 721: idx already equal to cbuf_len, so idx + 1 is definitely out of range

apps/restconf/restconf_methods.c
```diff
@@ -581,6 +583,409 @@ api_data_write(clicon_handle h,
...
+            if (brace_count == 2) {
+                cbuf_append(json_simple_patch,(int)'[');
+            }
+        }
+        cbuf_append(json_simple_patch,(int)c);
+    }
+    cbuf* json_simple_patch_2 = NULL;
+
+    // Insert a ']' before the last '}' to get the JSON to match what api_data_post() expects
+    for (int l = cbuf_len(json_simple_patch); l>= 0; l--) {
```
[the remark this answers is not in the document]
alanyanigersiklu: Yes, I changed it here and in the funtion yang_patch_strip_after_last_slash()

apps/restconf/restconf_methods.c
```diff
@@ -581,6 +583,409 @@ api_data_write(clicon_handle h,
...
+    xml2json_cbuf(cb, x_simple_patch, 1);
+
+    // Insert a '[' after the first '{' to get the JSON to match what api_data_post/write() expect
+    char *json_simple_patch_tmp = cbuf_get(cb);
+    int brace_count = 0;
+    for (int l = 0; l < strlen(json_simple_patch_tmp); l++) {
+        char c = json_simple_patch_tmp[l];
+        if (c == '{') {
+            brace_count++;
+            if (brace_count == 2) {
```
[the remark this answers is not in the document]
alanyanigersiklu: I first add the '[', and only then add the second brace. I've added a comment for clarity.

apps/restconf/restconf_methods.c
alanyanigersiklu: idx is not definitely out of range, since it gets a value in the loop after the initialization to cbuf_len. In the case that there is no slash, idx will not get a value in the loop, and will be out-of-range - which is a condition that I check in cbuf_trunc. However, I modified the code so that the out of range check is also done before calling cbuf_trunc().

apps/restconf/restconf_methods.c
```diff
@@ -581,6 +583,409 @@ api_data_write(clicon_handle h,
...
+    )
+{
+    cxobj **vec = NULL;
+    size_t veclen = 0;
+    char* tmp_val = NULL;
+    int ret = xpath_vec(xn, nsc, "%s", &vec, &veclen, key);
+    if (ret < 0) {
+        return ret;
+    }
+    for (int j = 0; j < veclen; j++) {
```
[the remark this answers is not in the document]
alanyanigersiklu: Actually, there shouldn't be a loop here at all, since veclen should always be 1. I've removed the loop, and done a simple assignment.

vsharapov-udelphi (approved):

alanyanigersiklu:
Hi Olof,

I've locally merged your master branch from https://github.com/clicon/clixon into my fork at  ***@***.***:alanyanigersiklu/clixon.git , branch, yang-patch-alanfork, without pushing it to the remote repo yet.
I'm getting a build error:

gcc -I. -I../../lib/src -I../../lib -I../../include -I../..  -DHAVE_CONFIG_H  -fPIC -D__PROGRAM__=\"clixon_backend\" -O2 -Wall  -c backend_plugin_restconf.c
backend_plugin_restconf.c: In function ‘restconf_pseudo_process_control’:
backend_plugin_restconf.c:266:13: error: ‘CLIXON_CONFIG_SBINDIR’ undeclared (first use in this function); did you mean ‘CLIXON_CONF_NS’?
  if ((dir = CLIXON_CONFIG_SBINDIR) == NULL){
             ^~~~~~~~~~~~~~~~~~~~~
             CLIXON_CONF_NS
backend_plugin_restconf.c:266:13: note: each undeclared identifier is reported only once for each function it appears in
Makefile:155: recipe for target 'backend_plugin_restconf.o' failed

Is there some #define that I have to change?

Thanks,
Alan
________________________________

apps/restconf/restconf_methods.c
```diff
@@ -582,68 +584,34 @@ api_data_write(clicon_handle h,
 
 #ifdef YANG_PATCH
 
-char * init_str() 
+/*! Free memory after a NULL pointer check
+  *
+  * @param [in] str     void pointer to memory to be freed
+  *
+  */
+static void yang_patch_free_mem(void *p)
```
olofhagsand, line 592: Just keeping with clixon style, please use:
```
static void
yang_patch_free_mem(void *p)
```

apps/restconf/restconf_methods.c
```diff
@@ -653,120 +621,223 @@ int get_xval(
         cxobj *xn = vec[j];
```
olofhagsand: It doesnt show in this diff, but, I think you can replace:
```
    if (veclen == 1) { //veclen should always be 1
        xn_tmp = vec[0];
    }
    if (xn_tmp != NULL) {
```
with
```
    if (veclen == 1) { //veclen should always be 1
       xn_tmp = vec[0];
```

apps/restconf/restconf_methods.c
```diff
@@ -653,120 +621,223 @@ int get_xval(
...
+  * @retval new cbuf with the modified json
+  * @retval NULL   Error
+  */
+
+static cbuf* yang_patch_xml2json_modified_cbuf(cxobj* x_simple_patch)
+{
+    cbuf *json_simple_patch = cbuf_new();
+    if (json_simple_patch == NULL)
+        return NULL;
+    cbuf* cb = cbuf_new();
```
olofhagsand, line 669: Clixon style: please move variable declarations to top of function, ie cb, json_simple_patch_tmp, brace_count.
Probably more cases.
Its OK to add them in a new body {}, but I get confused when they appear after code lines.
I know it is OK in current C, but my personal view is it is difficult to see what appears on the stack.

olofhagsand, line 669: Please check for (cb == NULL) and return NULL on failure. Also set error reason to catch memory errors. The typical way is to do something like:
```
if ((cb = cbuf_new()) == NULL){
  clicon_err(OE_UNIX, errno, "cbuf_new"); 
  return NULL;
}
```
 Please see all cases where a cbuf is created in the file, there seems to several instances of this.
The reason is that in the case of memory errors we want to catch it as close to the original fault as possible.

apps/restconf/restconf_methods.c
```diff
@@ -653,120 +621,223 @@ int get_xval(
...
+ * @param[in]  qvec      Vector of query string (QUERY_STRING)
+ * @param[in]  pretty    Set to 1 for pretty-printed xml/json output
+ * @param[in]  media_out Output media
+ * @param[in]  ds       0 if "data" resource, 1 if rfc8527 "ds" resource
+ * @param[in]  simplepatch_request_uri URI for patch request, e.g. "/restconf/data/ietf-interfaces:interfaces"
+ * @param[in]  target_val       value in "target" field of edit in YANG patch
+ * @param[in]  value_vec_len    number of elements in the "value" array of an edit in YANG patch
+ * @param[in]  value_vec        pointer to the "value" array of an edit in YANG patch
+ * @param[in]  x_simple_patch   pointer to XML containing module name, e.g. <ietf-interfaces:interface/>
+ */
```
olofhagsand, line 747: Please add documentation on retval for this function: -1, 0, 1?

olofhagsand (changes requested):
apps/restconf/restconf_methods.h
+#define TEMP_STR_MALLOC_SIZE 5000
can be removed.

Also, still lacking (1) XML support and (2) some basic testcases in test/test_restconf_patch.sh
(1) is more important, we can add 
(2) afterwards (but before it is enabled by default).

apps/restconf/restconf_methods.c
alanyanigersiklu: I changed the code to use cbufs, as you requested, but Olof wrote me today after reviewing my code that he prefers malloc functions to cbuf. In his words, "it seems the use of cbufs make the code more complex. Cbufs were not meant as a replacement of string functions (eg strcmp etc), just for buffer handling." So how should I handle this?

olofhagsand: Sorry, no that was not really what I said, or at least not what I meant. 
I support the use of cbufs, and not defining a string-malloc-lib specific to the file.
My concern is that the buffer code seems overly complex, and was asking whether there is a possibility to make it simpler?
One problem is it seems geared towards JSON, but it need to handle XML as well.
One could try to work with the internal xml structure: cxobj:s and the translate that to XML or JSON, instead.

olofhagsand (approved):

PR #269, opened 2021-09-30
Fixed CFLags for debug mode

s-bauer:
Fixing the behavior when using the `--enable-debug` flag -> the `CFLAGS` variable get's overwritten. Also logging out the `LDFLAGS` variable for debugging purposes.


olofhagsand:
Can you please add this in the documentation, eg as a PR here:
`https://clixon-docs.readthedocs.io/en/latest/install.html#configure-options`

PR #271, opened 2021-09-30
Added --without-sigaction option

s-bauer:
Even though QNX support the sigaction function, it does not support the `SA_RESTART` mode. The additional flag in the configure script allows passing `--without-sigaction` to disable the sigaction logic.


olofhagsand:
Can you please add this in the documentation, eg as a PR here:
`https://clixon-docs.readthedocs.io/en/latest/install.html#configure-options`

PR #272, opened 2021-09-30
Added QNX specific imports

s-bauer:
Fixed a few import issues with QNX. Is there a better way of conditionally configuring the imports? Maybe with autotools?

### DIFF
--- a/lib/clixon/clixon_stream.h
+++ b/lib/clixon/clixon_stream.h
@@ -58,6 +58,10 @@
  */
 #define EVENT_RFC5277_NAMESPACE "urn:ietf:params:xml:ns:netmod:notification"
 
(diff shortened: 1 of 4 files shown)


olofhagsand:
If possible, it would be better to make a generic check in autoconf, such as using eg:
https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Generic-Headers.html#index-AC_005fCHECK_005fHEADER-580
The reason is that there may be other OS:s with the same location of where include files are. Once you start putting ifdef:s for specific OS:s, as the number of OS:s increase so will the complexity of ifdef:s.

olofhagsand:
Closing due to old age

PR #290, opened 2021-11-30
Ability to not show entire modules in cli by defining them in config file

shmuelnatan:
Let's say that we have the following yang files module1.yang and module2.yang.
`module module1 {
....
}`
`module module2 {
....
}`

If one doesn't want these modules to show in cli then he needs to define them in the config file as following 
`<CLICON_CLI_SHOW_EXCLUDE>module1 module2</CLICON_CLI_SHOW_EXCLUDE>`

Modules can be separated by space in config file.

### DIFF
--- a/apps/cli/cli_show.c
+++ b/apps/cli/cli_show.c
@@ -72,6 +72,15 @@
 #include "clixon_cli_api.h"
 #include "cli_common.h" /* internal functions */
 
+/* Finds module name of cxobj 
+ */
+static char* get_module_name_from_cxobj(cxobj* x) {
+    while (xml_spec(x) == NULL) {
+        x = xml_parent(x);
+    }
+    return yang_argument_get(ys_module(xml_spec(x)));
+}
+
 /*! Given an xpath encoded in a cbuf, append a second xpath into the first
  *
  * The method reuses prefixes from xpath1 if they exist, otherwise the module prefix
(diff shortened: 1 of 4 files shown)


olofhagsand:
Question: instead of filtering on each sub-level (ie calling get_module_name_from_cxobj) why not just filter out on the top-level? I.e. in yang2cli().
I can see this if you want to filter out all augments that inherit the namespace from the original module. Is that the reason or is there something else?

shmuelnatan:
> Question: instead of filtering on each sub-level (ie calling get_module_name_from_cxobj) why not just filter out on the top-level? I.e. in yang2cli(). I can see this if you want to filter out all augments that inherit the namespace from the original module. Is that the reason or is there something else?

I tried deleting modules on the top level (In yang2cli_yspec() function) for the show tree, but the program kept on crashing. Your suggestion makes sense but I don't know how to do this without causing a crash.

olofhagsand:
OK, let me come up with a patch proposal and see if it works for you. Hopefully such a solution would be less intrusive.

dima1308:
@olofhagsand may I ask you what the status of it?
Do you plan to send a patch proposal or do we want us to continue with this patch?

olofhagsand:
Ah yes, 
The reason I have not come back with a patch is I am currently working with a new API in this area, with the aim to be done in clixon 5.5.
It does not seem like a good idea to make a patch to 5.4 as I originally planned but instead ensure that this is possible in Clixon 5.5 with the new API.
Thanks for reminding

olofhagsand:
Closing this since Clixon 5.5 introduced a new API making this PR obsolete.
See https://clixon-docs.readthedocs.io/en/latest/cli.html#autocli

PR #310, opened 2022-02-25
Convert int64, uint64 and decimal64 to string in xml to json

lukon89:
Convert int64, uint64 and decimal64 to string in xml to json convertion (regarding to RFC7951):

`A value of the "int64", "uint64", or "decimal64" type is represented
   as a JSON string whose content is the lexical representation of the
   corresponding YANG type as specified in Sections [9.2.1](https://datatracker.ietf.org/doc/html/rfc7951#section-9.2.1) and [9.3.1](https://datatracker.ietf.org/doc/html/rfc7951#section-9.3.1) of
   [[RFC7950](https://datatracker.ietf.org/doc/html/rfc7950)].`

### DIFF
--- a/lib/src/clixon_json.c
+++ b/lib/src/clixon_json.c
@@ -587,15 +587,25 @@ xml2json_encode_leafs(cxobj     *xb,
 	    else
 		cprintf(cb, "%s", body);
 	    break;
+	case CGV_INT64:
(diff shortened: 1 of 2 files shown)


olofhagsand:
Thanks for making Clixon more compliant

PR #328, opened 2022-05-02
Add CORS headers

dcornejo:
Insert CORS headers into replies. These are inserted unconditionally
and with fixed values. This is only tested against Chrome Javascript
and curl.


olofhagsand:
Please correct this compile error (see CI build ^)
```
 gcc -I. -I../../lib/src -I../../lib -I../../include -I../..  -D__PROGRAM__=\"clixon_restconf\" -I/clixon/build/include -DHAVE_CONFIG_H  -fPIC -O2 -Wall  -c restconf_api_native.c
restconf_api_native.c: In function 'restconf_reply_send':
restconf_api_native.c:140:31: error: 'req' undeclared (first use in this function); did you mean 'req0'?
  140 |     if (restconf_reply_header(req, "Access-Control-Allow-Origin", "*") < 0)
      |                               ^~~
      |                               req0
restconf_api_native.c:140:31: note: each undeclared identifier is reported only once for each function it appears in
```
Further, please add a test, or describe how to make such a test involving:

1. Use case for "working", and "not working"
2. For which variants of restconf the function (and test) is applicable, ie native vs fcgi; tls vs plain; http/1 vs /2, or "all".
3. If not, please add compile-time or run-time option default false.

PR #383, opened 2022-11-22
add regex unicode match support

nowaits:
Hi olof,
    Currently yang regex not support unicode match, this patch will add the spoort. The unicode transform code is from [cjson](https://github.com/DaveGamble/cJSON/blob/b45f48e600671feade0b6bd65d1c69de7899f2be/cJSON.c#L652)


olofhagsand:
I will add a copyright statement from cjson

nowaits:
> I will add a copyright statement from cjson

The function utf16_literal_to_utf8 from cjson is not the original function, the param has been changed.

olofhagsand:
OK, in your view, what part is from cjson?
Is the function "derived from" cjson?
Also, I could consider replacing the clixon json parser with cjson, ie parsing using the cjson structures and then post-parsing, translate from cjson structx to cxobj structs. Would you think this is a good idea?

nowaits:
Clixon json parser may have some bug, using cjson can fix the potential bugs.

PR #389, opened 2022-12-20
Issue#384: with-defaults basic-mode is not honored on get-config

joc-se:
Basic mode retrieval now explicit, as specified in RFC6243.


olofhagsand (approved):
Looks good.
The failure in the last test depends on an updated openconfig repo on 20/12.

Issue #414, opened 2023-01-31
Cannot get child using XPATH


dima1308:
When I execute the following query:
```
<get>
  <filter type="subtree">
    <rb-tg-if:interfaces xmlns:rb-tg-if="http://siklu.com/yang/tg/interfaces"/>
  </filter>
</get>
```
I get the following response:
```
<?xml version="1.0" encoding="utf-8"?>
<data xmlns="urn:ietf:params:xml:ns:netconf:base:1.0">
  <interfaces xmlns="http://siklu.com/yang/tg/interfaces">
    <host>
      <state>
        <mac-address>00:24:a4:16:51:a1</mac-address>
        <counters>
          <in-octets>157432</in-octets>
          <in-pkts>3730</in-pkts>
          <in-discards>0</in-discards>
          <in-errors>0</in-errors>
          <in-no-rule-discards>4</in-no-rule-discards>
          <out-octets>6791986</out-octets>
          <out-pkts>60877</out-pkts>
          <out-discards>0</out-discards>
          <out-errors>0</out-errors>
        </counters>
      </state>
    </host>
    <ports>
      <name>eth1</name>
      <alias>myeth1</alias>
      <admin-status>up</admin-status>
      <copper-specific-config>
        <auto-negotiate>true</auto-negotiate>
      </copper-specific-config>
      <connector-type>RJ-45</connector-type>
      <port-speed>10Gbps</port-speed>
      <state>
        <oper-status>up</oper-status>
        <actual-duplex-mode>full</actual-duplex-mode>
        <actual-port-speed>1Gbps</actual-port-speed>
        <counters>
          <in-octets>6407339</in-octets>
          <in-pkts>55797</in-pkts>
          <in-discards>0</in-discards>
          <in-errors>0</in-errors>
          <in-no-rule-discards>5</in-no-rule-discards>
          <out-octets>532508</out-octets>
          <out-pkts>8699</out-pkts>
          <out-discards>0</out-discards>
          <out-errors>0</out-errors>
        </counters>
      </state>
    </ports>
    <ports>
      <name>eth2</name>
      <alias>myeth2</alias>
      <admin-status>up</admin-status>
      <copper-specific-config>
        <auto-negotiate>true</auto-negotiate>
        <pse-out>true</pse-out>
      </copper-specific-config>
      <connector-type>RJ-45</connector-type>
      <port-speed>1Gbps</port-speed>
      <state>
        <oper-status>down</oper-status>
        <counters>
[...]

Issue #399, opened 2023-01-11
configuration of SNMP table indexes


dima1308:
Hi @olofhagsand 

Today, when we use the same yang container for both: regular use and SNMP table implementation, CLIXON assumes that the key leaf will be used as an index in the SNMP table. In most cases, it is OK, but in some cases, we may need to use a non-key leave as an SNMP table index.

We implemented [RFC 8348](https://datatracker.ietf.org/doc/rfc8348/) and we plan to add support for ENTITY-MIB.
As I understand, RFC 8348 (the standard Yang model for Hardware Management) defines how to map between Yang leaves and ENTITY-MIB OIDs:

```
   +--------------------------------+----------------------------------+
   | YANG data node in              | ENTITY-MIB object                |
   | /hardware/component            |                                  |
   +--------------------------------+----------------------------------+
   | name                           | entPhysicalName                  |
   | class                          | entPhysicalClass                 |
   |                                | entPhysicalVendorType            |
   | physical-index                 | entPhysicalIndex                 |
   | description                    | entPhysicalDescr                 |
   | parent                         | entPhysicalContainedIn           |
   | parent-rel-pos                 | entPhysicalParentRelPos          |
   | contains-child                 | entPhysicalChildIndex            |
   | hardware-rev                   | entPhysicalHardwareRev           |
   | firmware-rev                   | entPhysicalFirmwareRev           |
   | software-rev                   | entPhysicalSoftwareRev           |
   | serial-num                     | entPhysicalSerialNum             |
   | mfg-name                       | entPhysicalMfgName               |
   | model-name                     | entPhysicalModelName             |
   | alias                          | entPhysicalAlias                 |
   | asset-id                       | entPhysicalAssetID               |
   | is-fru                         | entPhysicalIsFRU                 |
   | mfg-date                       | entPhysicalMfgDate               |
   | uri                            | entPhysicalUris                  |
   | uuid                           | entPhysicalUUID                  |
   +--------------------------------+----------------------------------+

              YANG Data Nodes and Related ENTITY-MIB Objects
```

In the RFC 8348 Yang model, the name is defined as a key.
In the ENTITY-MIB, the entPhysicalIndex is defined as an index.

May I ask you to add an additional yang extension that will allow configuring which Yang leaves should be used as an index(ex) in the SNMP tables?

PR #415, opened 2023-01-31
Add support for custom snmp-keys

shmuelhazan:
This commit adds support for defining a
different table/list keys for snmp and
yang lists.

The definition is being done using the
cx-snmp:table-key extension.

Solves #399

Note: unit-tests currently fail due to https://github.com/clicon/clixon/issues/414.


dima1308:
@shmuelhazan may I ask you tp rebase this PR and resolve conflicts? 
I think, now the PR can work - the blocker bug is probably solved

PR #430, opened 2023-04-30
Support external yang features

shmuelhazan:
This PR adds support for using `if-feature` on features from imported yang models.


lib/src/clixon_yang.c
```diff
@@ -2610,7 +2610,7 @@ if_feature(yang_stmt    *yspec,
  * @param[in] ys   Feature yang statement to populate.
  * Bootstrapping: A feature is enabled if found in clixon-config
  */
-static int
+int
 ys_populate_feature(clicon_handle h,
```
shmuelhazan, line 2614: looks like I forgot to add it to the .h file.

PR #436, opened 2023-07-03
Fix cligen dir

rbgarga:
The main purpose of these commits is to fix build when `CLIGEN_DIR` is set or `--with-cligen=dir` is passed to configure.  Before this change CFLAGS and LDFLAGS were not respecting `$includedir` and `$libdir`

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -179,15 +179,14 @@ SH_SUFFIX=".so"
 LIBSTATIC_SUFFIX=".a"
 
 # This is for cligen
-AC_ARG_WITH([cligen], AS_HELP_STRING([--with-cligen=dir],[Use CLIGEN installation in this dir]))
+AC_ARG_WITH([cligen], [AS_HELP_STRING([--with-cligen=dir], [Use CLIGEN installation in this dir])], [
+  CLIGEN_DIR="$withval"
+])
+AC_SUBST(CLIGEN_DIR)
 if test -n "${CLIGEN_DIR}" -a -d "${CLIGEN_DIR}"; then
-        with_cligen=${CLIGEN_DIR}
-fi
-if test -d "${with_cligen}"; then
-  echo "Using CLIGEN here: ${with_cligen}"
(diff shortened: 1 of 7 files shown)


configure.ac
```diff
@@ -171,15 +171,14 @@ SH_SUFFIX=".so"
...
-        with_cligen=${CLIGEN_DIR}
-fi
-if test -d "${with_cligen}"; then
-  echo "Using CLIGEN here: ${with_cligen}"
-  CPPFLAGS="-I${with_cligen}/include ${CPPFLAGS}"
-  LDFLAGS="-L${with_cligen}/lib ${LDFLAGS}"
-  test -d "$with_cligen" && CLIGEN_PREFIX="$with_cligen"
+    echo "Using CLIGEN here: ${CLIGEN_DIR}"
+    CPPFLAGS="-I${CLIGEN_DIR}$(eval echo ${includedir}) ${CPPFLAGS}"
+    LDFLAGS="-L${CLIGEN_DIR}$(eval echo ${libdir}) ${LDFLAGS}"
```
olofhagsand, line 181: I have problems with 180-181 in the docker/test 2-step build. There cligen is installed in /clixon/build and clixon prefix is also /clixon/build. This causes those lines to evaluate to: "/clixon/build/clixo/build" which does not make sense.
It works if I instead set them to:
+    CPPFLAGS="-I$(eval echo ${includedir}) ${CPPFLAGS}"
+    LDFLAGS="-L$(eval echo ${libdir}) ${LDFLAGS}"
or to
+    CPPFLAGS="-I${CLIGEN_DIR} ${CPPFLAGS}"
+    LDFLAGS="-L$${CLIGEN_DIR} ${LDFLAGS}"
but then the original intent may be lost?
But I find it strange to concatenate the prefix of the package at hand (clixon) with the prefix of the package being looked for (cligen) which causes this problem.
If this could be sorted out better,  I am open for a discussion of intention and other ways to solve the 2-step docker build.

PR #437, opened 2023-07-06
Implement restconf netns default

rbgarga:
This PR branch is based on PR #436 and cannot be merged before it.

The idea is to make it possible to change default restconf network namespace as a configure option.  Only 3 last commits belong to this one.

### DIFF
--- a/configure
+++ b/configure
@@ -755,6 +755,7 @@ enable_debug
 with_cligen
 enable_yang_patch
 enable_publish
+with_restconf_netns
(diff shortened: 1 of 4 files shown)


olofhagsand:
With this change RESTCONF_NETNS_DEFAULT becomes a configurable compile-time option, instead of a static.
My only comment I have is whether this option should be a runtime option?
If so it should be added to clixon-configure.yang instead.

Issue #439, opened 2023-08-14
CLICON_CONFIGDIR with external subsystems causes endless looping


pprindeville:
In the XML parsing in `parse_configfile_one()` we have the [following loop](https://github.com/clicon/clixon/blob/master/lib/src/clixon_options.c#L329-L342).

The problem with the loop is that you can't "ignore" the nodes by just doing a `continue`.  They need to be deleted from `xe` so that the next call to examine the head of the vector:

```
while ((xec = xml_child_i_type(xe, 0, CX_ELMNT)) != NULL) {
```

doesn't find them still there.  We need to do a `xml_child_rm()` to remove them from `xe` before continuing the loop.

pprindeville:
Fixed in PR #440.

pprindeville:
Fixed w/ the merge of PR #440.

PR #440, opened 2023-08-14
Properly ignore skipped nodes

pprindeville:
Fix for issue #439.

Also add matching debugs for starting/completing parsing config files.


lib/src/clixon_options.c
```diff
@@ -354,6 +357,9 @@ parse_configfile(clicon_handle  h,
                 /* Append to master (removed from xe) */
                 if (xml_addsub(xt, xec) < 0)
                     goto done;
+                continue;
+ignore:
```
olofhagsand, line 361: Please avoid goto:s unless it is to the single return, see:
https://github.com/clicon/clixon/blob/master/CONTRIBUTING.md#return-values

Also please extend the test https://github.com/clicon/clixon/blob/master/test/test_configdir.sh if possible with the usecase that causes a loop. In that way it is added to the regression.

olofhagsand (commented):
Please extend the test https://github.com/clicon/clixon/blob/master/test/test_configdir.sh if possible with the usecase that causes a loop. In that way it is added to the regression.

lib/src/clixon_options.c
pprindeville: Done

PR #442, opened 2023-08-15
Add autoconf handling of clixon user/group in config files

pprindeville:
Pass the user/group into `configure` so that the `.xml.in` file can get `AC_SUBST()`'d appropriately.


olofhagsand:
I am some problems with this. The user/group is in the config file and defined in the clixon config YANG: https://github.com/clicon/clixon/blob/master/yang/clixon/clixon-config%402023-05-01.yang :
CLICON_SOCK_GROUP / CLICON_BACKEND_USER / CLICON_RESTCONF_USER.
The default value of CLICON_SOCK_GROUP is "clicon" and CLICON_RESTCONF_USER is "www-data"
We can discuss that, maybe they should not have default values?
The autoconf change proposed in this PR affects only individual examples / applications of clixon.
In the clixon repo (here), it affect only the main example. It currently exists because some tests use it. I even plan to make it independent and then move it to clixon-examples. 
It could make sense to move the PR there, and to other clixon applications, such as clixon-openwrt/clixon-controller. But not in the core clixon module.
And the reason is that it is application-dependent. Having default values in the YANG may even be wrong. Adding an autoconf value on top of the default value is confusing.

PR #447, opened 2023-08-20
Make error output consistently sent to same destination

pprindeville:
_In progress_

Drop `perror()` and `fprintf()`, otherwise messages won't be seen if service is started from `systemd` or `procd`.

### DIFF
--- a/apps/backend/backend_main.c
+++ b/apps/backend/backend_main.c
@@ -1029,13 +1029,13 @@ main(int    argc,
        daemonized errors OK. Before this stage, errors are logged on stderr 
        also */
     if (foreground==0){
+        clicon_log_init(__PROGRAM__, dbg?LOG_DEBUG:LOG_INFO,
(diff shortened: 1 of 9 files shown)


test/sum.sh
```diff
@@ -36,6 +36,32 @@ if [ $# -gt 0 ]; then
     exit -1
 fi
 
+if [ -z "$(which clixon_util_xml)" ]; then
```
olofhagsand, line 39: This probably should be done in lib.sh
The reason is that "sum.sh" is just one way to run the tests, there are others, such as individually, ./mem.sh, ./all.sh

olofhagsand (changes requested):

test/sum.sh
pprindeville: Dropped.

pprindeville:
Handled as part of PR #449.

PR #459, opened 2023-10-23
Fix hardcoded paths

pprindeville:
Even though we can tell `./configure` where we want `clixon_netconf` installed, that doesn't help if `clixon_client.c` has the path hardcoded into it.  Same for `clixon_netconf_ssh_callhome.c`.


pprindeville:
Requires PR #460 first.

pprindeville:
#459 requires #460.

pprindeville:
Requires #464.

PR #460, opened 2023-10-23
Avoid additional 'eval echo ...' sequences

pprindeville:
Simplifying the previous "fix --with-cligen build option" so that we don't have to keep using `eval echo $var`.

### DIFF
--- a/configure
+++ b/configure
@@ -664,11 +664,14 @@ ac_ct_CXX
 CXXFLAGS
 CXX
 CPP
+SYSCONFDIR
+SBINDIR
 LOCALSTATEDIR
 LIBEXECDIR
 LIBDIR
+INCLUDEDIR
 DATADIR
-SYSCONFDIR
+BINDIR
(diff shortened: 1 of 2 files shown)


pprindeville:
cc: @rcmcdonald91

olofhagsand:
This seems to be a part of https://github.com/clicon/clixon/pull/459 as well. Why is it broken out /why is it still in #459?

pprindeville:
> This seems to be a part of #459 as well. Why is it broken out /why is it still in #459?

Because #459 will rebase onto #460 and it will go away, but #459 requires it to get a reliable a reliable run.

olofhagsand:
This part of configure is there primarily on request for downstream build for netgate. Please ensures it works in that context and its OK.

pprindeville:
> This part of configure is there primarily on request for downstream build for netgate. Please ensures it works in that context and its OK.

Yes, tested it and it's working fine.  But Netgate is still on 6.3.0, aren't they?

PR #464, opened 2023-11-04
Revert PR#457 as it breaks cross-compilation

pprindeville:
Adding `-I${INCLUDEDIR}` unconditionally to `CPPFLAGS` will cause the compiler to search the build-host's local headers, which will be disastrous.


pprindeville:
cc: @rcmcdonald91 @rbgarga

PR #470, opened 2023-11-30
Fix issues with valgrind w/ sudo

pprindeville:
Explode string into list of words for `spawn`.


pprindeville:
`[split $clixon_cli " "]` might have been simpler but I couldn't get that working.

PR #476, opened 2024-01-11
Drop __FUNCTION__ from clixon_debug() calls

pprindeville:
Trying to copy what `clixon_log_fn()` does.

### DIFF
--- a/apps/backend/backend_client.c
+++ b/apps/backend/backend_client.c
@@ -146,7 +146,7 @@ ce_event_cb(clixon_handle h,
     struct client_entry *ce = (struct client_entry *)arg;
     cbuf                *cbce = NULL;
 
-    clixon_debug(CLIXON_DBG_CLIENT, "%s op:%d", __FUNCTION__, op);
+    clixon_debug(CLIXON_DBG_CLIENT, "op:%d", op);
     switch (op){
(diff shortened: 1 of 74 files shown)


pprindeville:
Hmm... it's going to complain in a lot of places that the format string is empty, even though that's valid C...

olofhagsand:
In what way complain?

pprindeville:
> In what way complain?

```
In file included from ../../lib/clixon/clixon.h:79,
                 from netconf_rpc.c:71:
netconf_rpc.c: In function ‘netconf_notification_cb’:
netconf_rpc.c:462:37: warning: zero-length gnu_printf format string [-Wformat-zero-length]
  462 |     clixon_debug(CLIXON_DBG_CLIENT, "");
      |                                     ^~
../../lib/clixon/clixon_debug.h:68:97: note: in definition of macro ‘clixon_debug’
   68 | #define clixon_debug(l, _fmt, args...) clixon_debug_fn(NULL, __FUNCTION__, __LINE__, (l), NULL, _fmt, ##args)
      |                                                                                                 ^~~~
```

pprindeville:
Okay, fixed the warnings by using `_Pragma()` expansions as part of the macro.

Issue #411, opened 2023-01-23
yang bits type is not SNMP supported


doron2020:
Hi, It looks like a bit type is not converted to a yang OID and snmpwalk will return error:

iso.3.6.1.4.1.31926.35.3.1.4.5.100.111.114.111.110 = OID: iso.3.6.1.4.1.31926.35.4
Error: OID not increasing: iso.3.6.1.4.1.31926.35.3.1.4.5.100.111.114.111.110

An example: 

leaf value-bitmap {
        type bits {
          bit value1 { position 0; }
          bit value2 { position 1; }
          bit value3 { position 2; }
          bit value4 { position 3; }
        }  
          config false;
          smiv2:max-access "read-only";
          smiv2:oid "1.3.6.1.4.1.31926.35.3.1.5";
}

smidump is converting a MIB "SYNTAX BITS" to a yang bits type.
Thanks

mico-micic:
Hi @doron2020 
I've just created a SNMP bits implementation. Check PR: https://github.com/clicon/clixon/pull/501

PR #491, opened 2024-02-08
SNMP bits support

mico-micic:
Add support for SNMP "bits" data type (Feature Request #411)

@olofhagsand: May I ask you to take a look at this pull request and give me some hints on how to improve it? Currently there are no unit tests, I will write them later. So my intention is not to merge this request now, but to get feedback on whether I am "on the right track".

Note: I need to define the mapping between YANG data type "bits" and ASN_OCTET_STR, even if there is an ASN_BITS data type. The reason is that I get the error message "Expected type:4, got: 3" in https://github.com/clicon/clixon/blob/master/apps/snmp/snmp_handler.c#L747 when I use the ASN_BITS type. As far as I understand, the expected type is specified by the nets-snmp library in `netsnmp_request_info`, so maybe there is a bug? (I have not checked it.)

Note 2: In all functions (like `type_xml2snmp_pre`), the data read or written to/from snmp is treated as a string. This does not work for bits as there can be multiple 0x00 bytes. For this reason I had to implement it directly in `type_xml2snmp`.

Thanks,
Mico

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,8 @@ jobs:
       run: (cd example; make)   
     - name: install example
       run: (cd example; sudo make install)
-  docker-alpine-tests:
+
+  docker-alpine-tests-a-j:
     runs-on: ubuntu-latest
     defaults:
       run:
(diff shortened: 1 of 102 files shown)


olofhagsand:
Please resolve the test_snmp_get.sh conflict

olofhagsand (changes requested):
I dont understanf the multiple Makefile.in patches?
It seems this diffs is reoccuring in multiple commits.
Can you please clean this up / squash and maybe separate into separate PR:s?

olofhagsand:
Can you please split this PR into two?
One for the new bit-manipulating functions in clixon-lib
And one for SNMP bit support?
Also, there are multiple commits that null each other, a squashing of multiple commits would be in order.
Style: I see some extra empty lines and occasional spaces before \n.

mico-micic:
PR was splittet in two. This one here contains the snmp support: https://github.com/clicon/clixon/pull/501.
(Closing this PR here..)

PR #493, opened 2024-02-12
Remove null terminator for SNMP octet string

mico-micic:
In the SNMP protocol all tranfered values (bindings) have a defined length. Therefore the octet string does not need a null terminator. If there is one, it is transfered as control character and may confuse SNMP agents like snmp4j. (snmpget shows a "." character at the end what ist not correct.)

Check the result without and with this fix ("." at the end of "Some string value")

```
--> Without this fix:
$ snmpget -v2c -c private localhost SIMPLE-MIB::testAlphaOneConfstring.0
SIMPLE-MIB::testAlphaOneConfstring.0 = STRING: Some string value.

--> With this fix:
$ snmpget -v2c -c private localhost SIMPLE-MIB::testAlphaOneConfstring.0
SIMPLE-MIB::testAlphaOneConfstring.0 = STRING: Some string value
```

### DIFF
--- a/apps/snmp/snmp_lib.c
+++ b/apps/snmp/snmp_lib.c
@@ -917,11 +917,12 @@ type_xml2snmp(char       *snmpstr,
         break;
     }
     case ASN_OCTET_STR: // 4
-        *snmplen = strlen(snmpstr)+1;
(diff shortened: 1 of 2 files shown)


olofhagsand:
Looks good to me

PR #515, opened 2024-05-02
Yang schema mount fixes

colinhusu:
While trying out clixon's Yang schema mount support, I discovered and fixed some issues. See the individual commit messages for details.

### DIFF
--- a/apps/cli/cli_auto.c
+++ b/apps/cli/cli_auto.c
@@ -119,6 +119,7 @@ cli_auto_edit(clixon_handle h,
 {
     int           retval = -1;
     char         *api_path_fmt;  /* xml key format */
+    char         *api_path_fmt01 = NULL;
     char         *api_path = NULL;
     char         *treename;
     pt_head      *ph;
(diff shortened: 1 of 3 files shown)


olofhagsand:
Good initiative, the yang schema mount as far as I know, is only used in the clixon controller project. 
It would be good with more usage.

colinhusu:
The "cli_dbxml(): Fix handling of mountpoints" patch had an error, it didn't handle the case where more than one api_path_fmt argument was passed in via argv, which made test_autocli_grouping.sh fail. I have adjusted the patch to properly handle that case.

> Good initiative, the yang schema mount as far as I know, is only used in the clixon controller project.
> It would be good with more usage.

FYI, we have a relatively simple use-case for yang schema mount right now. We mount an external yang module inside of an empty presence container from our own yang module. The yang module to be mounted is known at build time and not determined dynamically.

olofhagsand:
Looks good, merged, thanks.
> FYI, we have a relatively simple use-case for yang schema mount right now. We mount an external yang module inside of an empty presence container from our own yang module. The yang module to be mounted is known at build time and not determined dynamically.

Do you have more info available on your project / how it uses mount-points?

PR #519, opened 2024-05-14
RFC: Partly fix handling of mountpoints in the 'up' command

colinhusu:
The `up` command right now does not properly handle mountpoints. Partly fix that.

Note that this does not yet fix the case where the current level is the top-level of the mounted module. In that case the `up` command should switch to the presence container that holds the mounted module. Instead it switches all the way up to the topmost level, so it has the same effect as the `top` command.

Fixing that would be more involved: `co_up()` on the top-level `cg_obj` of a mounted module returns `NULL`. That case would need to be handled by finding the `cg_obj` matching the mount-point container, i.e. the path stored in `"cli-edit-mtpoint"`, and setting that as the new `cligen_ph_workpoint`. But I don't now how to get that cg_obj.

### DIFF
--- a/apps/cli/cli_auto.c
+++ b/apps/cli/cli_auto.c
@@ -248,12 +248,14 @@ cli_auto_up(clixon_handle h,
     cvec    *cvv1 = NULL; /* copy */
     char    *api_path_fmt0;  /* from */
     char    *api_path_fmt1;  /* to */
+    char    *api_path_fmt2 = NULL;  /* 'to' with mountpoint prepended */
     char    *api_path = NULL;
     int      i;
     int      j;
     size_t   len;
     cvec    *cvv_filter = NULL;
     yang_stmt *yspec0;
+    char    *mtpoint = "";
 
     if (cvec_len(argv) != 1){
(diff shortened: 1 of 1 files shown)


olofhagsand:
I know,there is a lot of  newautocli code in the controller, that needs to be retrofitted into more generic clixon code. But I never got edit modes to work, so the submission is appreciated.

colinhusu:
Thanks for merging this!

> But I never got edit modes to work, so the submission is appreciated.

Do you remember any specific problems in this area? For me edit mode basically appears to work in combination with mountpoints.

olofhagsand:
Operation across mountpoints, such as "up"

olofhagsand:
> Do you remember any specific problems in this area? For me edit mode basically appears to work in combination with mountpoints.

Not really but there is an issue in the controller project:
https://github.com/clicon/clixon-controller/issues/83

colinhusu:
I don't remember having seen error messages similar to the one in the mentioned clixon-controller issue.

However, I did get some error-messages before due to inconsistent cli-edit-* states caused by earlier calls to up and top commands. But these should be addressed now.

olofhagsand:
We will revisit that issue and see if it works now.

PR #520, opened 2024-05-22
Add run python3

khromenokroman:
This commit includes a new function `cli_start_python3` which executes Python3 scripts from within the CLI. The function operates as a new command in the CLI interface, accepting a path to the script as an argument. This enhancement improves the versatility of the Clixon CLI interface by allowing direct execution of Python scripts.

### DIFF
--- a/configure
+++ b/configure
@@ -5302,11 +5302,11 @@ if test x$ac_prog_cxx_stdcxx = xno
 then :
   { printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking for $CXX option to enable C++11 features" >&5
 printf %s "checking for $CXX option to enable C++11 features... " >&6; }
-if test ${ac_cv_prog_cxx_11+y}
+if test ${ac_cv_prog_cxx_cxx11+y}
 then :
   printf %s "(cached) " >&6
 else $as_nop
-  ac_cv_prog_cxx_11=no
+  ac_cv_prog_cxx_cxx11=no
 ac_save_CXX=$CXX
 cat confdefs.h - <<_ACEOF >conftest.$ac_ext
 /* end confdefs.h.  */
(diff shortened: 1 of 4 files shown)


krihal:
What are the advantage over adding a shebang for Python3 in the script and then use something like this in the CLI specification?

```
shell("System command"), cli_start_shell("bash");{
  <source:rest>("Single shell command"), cli_start_shell("bash");
}
```

Any Python script can then be called as long as `#!/bin/env python3` or similar are added to the top of the file and the file is executable and "shell" can of course be renamed.

```
debian@khn-dev /> shell /tmp/test.py
foo
```

configure
olofhagsand, line 1: This is a generated file and should most probably not be here (unless you changed configure.ac)

olofhagsand:
It would be nice to add a documentation section somewhere here: https://clixon-docs.readthedocs.io/en/latest/cli.html on how to use scripts (specifically python) as callbacks.

khromenokroman:
> What are the advantage over adding a shebang for Python3 in the script and then use something like this in the CLI specification?
> 
> ```
> shell("System command"), cli_start_shell("bash");{
>   <source:rest>("Single shell command"), cli_start_shell("bash");
> }
> ```
> 
> Any Python script can then be called as long as `#!/bin/env python3` or similar are added to the top of the file and the file is executable and "shell" can of course be renamed.
> 
> ```
> debian@khn-dev /> shell /tmp/test.py
> foo
> ```

That's right, you can do this. But then it will be like this
Снимок экрана от 2024-05-22 13-16-10

And with my version it will be like this
Снимок экрана от 2024-05-22 13-18-23

I think it's better this way. But I could be wrong :)

configure
khromenokroman: I agree, this is a mistake, I'm sorry

khromenokroman:
> It would be nice to add a documentation section somewhere here: https://clixon-docs.readthedocs.io/en/latest/cli.html on how to use scripts (specifically python) as callbacks.

I'll try to do it now

olofhagsand:
> > It would be nice to add a documentation section somewhere here: https://clixon-docs.readthedocs.io/en/latest/cli.html on how to use scripts (specifically python) as callbacks.
> 
> I'll try to do it now

You need to checkout the old version of configure for tests to parse

olofhagsand:
>  I think it's better this way. But I could be wrong :)
Yes, but
can you do it more generalized, so that it is not hardcoded to /usr/bin/python3? I can see it being useful for other scripts (or even if python is installed elsewhere.u
This is the advantage of the existing cli_start_shell.

olofhagsand:
I see you close this PR. I think it is important to improve the scripting interface, and also its documentation, and I think your comments made sense. Are you planning to make a new PR?

PR #522, opened 2024-05-23
Create python func

khromenokroman:
I fixed the code and took into account your comments

### DIFF
--- a/apps/cli/cli_common.c
+++ b/apps/cli/cli_common.c
@@ -774,6 +774,67 @@ cli_set_mode(clixon_handle h,
     return retval;
 }
 
+/**
+ * @brief Runs a Python script in a new process
+ *
+ * @param h Clicon handle
+ * @param cvv CLIgen variable vector
+ * @param argv function arguments
+ *
+ * The function creates a new child process in which the Python script is launched.
+ * In the child process, before launching the script, environment variables are set
+ * taken from cvv.
+ *
+ * The function returns the exit code of the Python script, or a value of (-1) on error.
+ *
+ * @return Returns 0 on success, -1 in case of error
+ */
+int cli_start_python3(clixon_handle h, cvec *cvv, cvec *argv) {
+    int pid = 0;
+    int status = 0;
+    char *script_path = NULL;
+    char *runner = "python3";
+    struct passwd *pw = NULL;
+
+    if (cvec_len(argv) > 1){
+        fprintf(stderr,"A lot of arguments");
+        return -1;
+    }
+
+    if (cvec_len(argv) == 1){
+        script_path = cv_string_get(cvec_i(argv, 0));
+    }
+
+    if (cvec_len(cvv) == 2){
+        script_path = cv_string_get(cvec_i(cvv, 1));
+    }
+
+    if ((cvec_len(cvv) == 2) && (cvec_len(argv) == 1)){
+        fprintf(stderr,"Both the arguments of the function and the vector of values are specified\n");
+        return -1;
+    }
+
+    if ((pw = getpwuid(getuid())) == NULL){
+        fprintf(stderr,"getpwuid -> %s", strerror(errno));
+        return -1;
+    }
+    if (chdir(pw->pw_dir) < 0){
+        fprintf(stderr,"chdir -> %s", strerror(errno));
+        return -1;
+    }
+
+    if ((pid = fork()) == 0) {
+        execlp(runner, runner, script_path, NULL);
+        perror("Error run script");
+        exit(0);
+    }
+
+    if (waitpid(pid, &status, 0) == pid)
+        return WEXITSTATUS(status);
+    else
+        return -1;
+}
+
 /*! Start bash from cli callback
  *
  * Typical usage: shell("System Bash") <source:rest>, cli_start_shell();
--- a/apps/cli/cli_common.c
+++ b/apps/cli/cli_common.c
@@ -803,8 +803,8 @@ cli_set_mode(clixon_handle h,
  */
 int
 cli_start_program(clixon_handle h,
-                      cvec         *cvv,
-                      cvec         *argv)
+                  cvec         *cvv,
+                  cvec         *argv)
 {
     int pid = 0;
     int retval = -1;
--- a/apps/cli/cli_common.c
+++ b/apps/cli/cli_common.c
@@ -60,6 +60,7 @@
 #include <sys/param.h>
 #include <sys/mount.h>
 #include <pwd.h>
+#include <libgen.h>
 
 /* cligen */
 #include <cligen/cligen.h>
@@ -810,6 +811,7 @@ cli_start_program(clixon_handle h,
     char *script_path = NULL;
     char *runner = NULL;
     char *buf = NULL;
+    char *work_dir = NULL, *reserve_path = NULL;
     struct passwd pw, *pwresult = NULL;
     size_t bufsize;
     int s;
@@ -839,7 +841,10 @@ cli_start_program(clixon_handle h,
     if (cvec_len(cvv) == 2){
         script_path = cv_string_get(cvec_i(cvv, 1));
     }
-
+    if (script_path){
+        reserve_path = strdup(script_path);
+        work_dir = dirname(reserve_path);
+    }
     bufsize = sysconf(_SC_GETPW_R_SIZE_MAX);
     if (bufsize == -1){
         bufsize = 16384;
@@ -862,7 +867,7 @@ cli_start_program(clixon_handle h,
     /* main run */
     if ((pid = fork()) == 0) {
         /* child process */
-        if (chdir(pw.pw_dir) < 0){
+        if ((work_dir ? chdir(work_dir) : chdir(pw.pw_dir)) < 0) {
             clixon_err(OE_PLUGIN, errno, "chdir");
         }
         execlp(runner, runner, script_path, NULL);
@@ -884,6 +889,8 @@ cli_start_program(clixon_handle h,
     done:
         if(buf)
             free(buf);
+        if(reserve_path)
+            free(reserve_path);
         return retval;
 }
 
--- a/apps/cli/cli_common.c
+++ b/apps/cli/cli_common.c
@@ -774,31 +774,34 @@ cli_set_mode(clixon_handle h,
     return retval;
 }
 
-/**
+/*!
  * @brief Runs a Python script in a new process
  *
- * @param h Clicon handle
- * @param cvv CLIgen variable vector
- * @param argv function arguments
- *
+ * @param[in] h Clicon handle
+ * @param[in] cvv CLIgen variable vector
+ * @param[in] argv function arguments
+ * @retval     0    OK
+ * @retval    -1    Error
  * The function creates a new child process in which the Python script is launched.
  * In the child process, before launching the script, environment variables are set
  * taken from cvv.
  *
  * The function returns the exit code of the Python script, or a value of (-1) on error.
- *
- * @return Returns 0 on success, -1 in case of error
  */
-int cli_start_python3(clixon_handle h, cvec *cvv, cvec *argv) {
+int
+cli_start_python3(clixon_handle h,
+                      cvec         *cvv,
+                      cvec         *argv)
+{
     int pid = 0;
-    int status = 0;
+    int retval = -1;
     char *script_path = NULL;
     char *runner = "python3";
     struct passwd *pw = NULL;
 
     if (cvec_len(argv) > 1){
-        fprintf(stderr,"A lot of arguments");
-        return -1;
+        clixon_err(OE_PLUGIN, EINVAL, "A lot of arguments");
+        goto done;
     }
 
     if (cvec_len(argv) == 1){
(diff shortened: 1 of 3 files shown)


apps/cli/cli_common.c
```diff
@@ -774,6 +774,67 @@ cli_set_mode(clixon_handle h,
...
+ */
+int cli_start_python3(clixon_handle h, cvec *cvv, cvec *argv) {
+    int pid = 0;
+    int status = 0;
+    char *script_path = NULL;
+    char *runner = "python3";
+    struct passwd *pw = NULL;
+
+    if (cvec_len(argv) > 1){
+        fprintf(stderr,"A lot of arguments");
```
olofhagsand, line 800: Please look at CONTRIBUTING.md regarding style. Use clixon_err for errors and preferably a done: clasue instead of direct returns.

olofhagsand (changes requested):
A restconf test also fails, it seems unrelated to this, but need to check.

apps/cli/cli_common.c
```diff
@@ -774,6 +774,73 @@ cli_set_mode(clixon_handle h,
...
+ * @retval     0    OK
+ * @retval    -1    Error
+ * The function creates a new child process in which the Python script is launched.
+ * In the child process, before launching the script, environment variables are set
+ * taken from cvv.
+ *
+ * The function returns the exit code of the Python script, or a value of (-1) on error.
+ */
+int
+cli_start_python3(clixon_handle h,
```
dima1308, line 792: May I ask what is the reason for this feature?
It can be a security problem when you allow to execute an arbitrary Python script.
In any case, the decision of which runner to use is normally defined by the script's shebang. Maybe no need to define the runner name (`python3` in this case). Instead, just pass the script name to the exec() call. In this case, the function will allow running any executable (not just a Python script).

khromenokroman: That's the point, the administrator can specify a script in the cli that can be executed and if something goes wrong, then we will return to the cli, and if we do system(), then bash starts, and then the script, and if something goes wrong, then we get into bash, this is a security problem It's not in my solution. it is also assumed that the user is only in the cli and does not have access to the system, but maybe I am outdated and something has changed :)

dima1308: @khromenokroman 
In any case, why not rely on Script's shebang string?

khromenokroman: Sorry, I took into account all the comments and changed the code, but left the description the same, now I have made changes to the description of the function

apps/cli/cli_common.c
```diff
@@ -774,6 +774,67 @@ cli_set_mode(clixon_handle h,
...
+ *
+ * The function returns the exit code of the Python script, or a value of (-1) on error.
+ *
+ * @return Returns 0 on success, -1 in case of error
+ */
+int cli_start_python3(clixon_handle h, cvec *cvv, cvec *argv) {
+    int pid = 0;
+    int status = 0;
+    char *script_path = NULL;
+    char *runner = "python3";
```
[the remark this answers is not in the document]
khromenokroman: I agree, I have made changes

khromenokroman:
@olofhagsand Made the function more versatile

khromenokroman:
@olofhagsand I don't really understand the project yet, I wanted to clarify, I need to look at why the Error in test failed test_restconf_callhome.sh errcode=255 and Error in test_restconf_ssl_certs.sh errcode=255 I have not made any changes to the restconf functionality

khromenokroman:
@olofhagsand When I check in, I'm fine
Снимок экрана от 2024-05-27 16-13-31
I did it on my computer according to the documentation https://github.com/clicon/clixon/blob/master/test/README.md Getting started

apps/cli/cli_common.c
```diff
@@ -774,6 +774,94 @@ cli_set_mode(clixon_handle h,
...
+    }
+    if (cvec_len(cvv) == 2){
+        script_path = cv_string_get(cvec_i(cvv, 1));
+    }
+
+    if ((pw = getpwuid(getuid())) == NULL){
+        clixon_err(OE_PLUGIN, errno, "getpwuid");
+        goto done;
+    }
+    if (chdir(pw->pw_dir) < 0){
```
dima1308, line 844: Here you are changing the directory for the whole process. It may impact other services that runs on the same process and rely on the current working directory.
It is better to do the following:
1. fork() the process
2. for the child process:
  a. change the current directory of the child process (it will leave the current directory of the main process unchanged)
  b. execute the script you want
3. for the parent process - wait for the child process to be finished

Other notes: 
1. In case of failure, you execute `exit(0)` - this will terminate the whole process including other services that may run in this process. It will be better to report an error and let the upper layer decide how to handle errors. 
2. It is better to pass the directory from where to execute the `runner` as a separate argument. It will allow the caller to decide what is the current directory from where you run the script. (If the caller does not specify the directory, you can use the initial working directory by default). This is important especially if you plan to use Python scripts that may require to be executed from a specific directory that is different from the initial working directory - this may allow the Python interpreter to find its dependencies.

olofhagsand:
> @olofhagsand I don't really understand the project yet, I wanted to clarify, I need to look at why the Error in test failed test_restconf_callhome.sh errcode=255 and Error in test_restconf_ssl_certs.sh errcode=255 I have not made any changes to the restconf functionality

No, its something with the regression test context, maybe alpine libssl or curl upgraded to something where some ssl operations fail. Looking into it, not related to this, other tests fail as well.

apps/cli/cli_common.c
khromenokroman: done

khromenokroman:
@olofhagsand I made changes, but I can't somehow run CI at home to check the correctness of the changes made, and not to bother you

olofhagsand:
> @olofhagsand I made changes, but I can't somehow run CI at home to check the correctness of the changes made, and not to bother you
Just run 'make test' on the top-level.

khromenokroman:
> > @olofhagsand I made changes, but I can't somehow run CI at home to check the correctness of the changes made, and not to bother you
> > Just run 'make test' on the top-level.

Thanks

apps/cli/cli_common.c
```diff
@@ -774,6 +775,125 @@ cli_set_mode(clixon_handle h,
...
+ * @param[in] h Clixon handle
+ * @param[in] cvv Vector of command variables
+ * @param[in] argv Function arguments
+ * @retval   0 OK, returns the exit code of the program
+ * @retval  -1 Error
+ *
+ */
+int
+cli_start_program(clixon_handle h,
+                      cvec         *cvv,
```
olofhagsand, line 806: Please follow https://github.com/clicon/clixon/blob/master/CONTRIBUTING.md#c-style regarding parameters and local variables indentation.

example/main/example_cli.cli
```diff
@@ -71,6 +71,14 @@ debug("Debugging parts of the system"){
 shell("System command"), cli_start_shell("bash");{
   <source:rest>("Single shell command"), cli_start_shell("bash");
 }
+
+run_program_err("Run program"), cli_start_program();
```
olofhagsand, line 75: The primary purpose of the example application is for the test-suite (under test/). 
The present CLI test command is referenced in test/test_cli.sh for example.
Other applications are in the separate repo https://github.com/clicon/clixon-examples
In my view this is primarily a "user-guide" example, ie how to run a python script from the CLI.
Therefore, if there is no regression test for this, the added commands do not belong here.
I think the best place for them are adding them to the CLI section of the user-guide instead, describing how a script could be started from the CLI.
The place to add such a subsection would be here: https://github.com/clicon/clixon-docs/blob/master/cli.rst
That said, if this command can replace the existing cli_start_shell command, one could replace it, and also change the test case in test_cli.sh so that it runs this new command instead.

To add such a section,

olofhagsand (commented):
@dima1308 added good comments on security. 
Initially the start shell command was just a sketch, but as this improvement is more elaborate, maybe one should add security considerations as well.
First, note that these commands run in the user context of the CLI not the backend, so the danger of executing random programs from root is not applicable.
But it still opens a way to run a random program from the CLI, ie as a user of the CLI which is probably not a good idea for a product.
One way to address this would be to create a chroot() after forking where such scripts could be accessed and run. 
At the very least these security considerations should be highlighted in (the proposed) sub-section of the user-guide.
I am also somewhat reluctant to add this in the default example (also applies tot he original function). As I said, the original cli_start_shell() function was added as a sketch, but I see that many users take the example application as-is, so there is a danger to having it in the code.
Maybe one could make the limitation to replacing cli_start_shell in cli_common.c with this function, document it properly in the user guide, but remove references completely from the clispec application (example_cli.cli).
In that way, a new user would have to add it explicitly (with appropriate caveats) and not get the command bundled implicitly.

apps/cli/cli_common.c
khromenokroman: done

khromenokroman: > Please follow https://github.com/clicon/clixon/blob/master/CONTRIBUTING.md#c-style regarding parameters and local variables indentation.

done

khromenokroman:
> But it still opens a way to run a random program from the CLI, ie as a user of the CLI which is probably not a good idea for a product.

and how can the user run any program, the cli-spec is available only to the administrator, only he can add a call to which script there, the user can only select the pre-defined scripts that the administrator has set, or do I not understand something?

example/main/example_cli.cli
khromenokroman: remove in examle.cli

apps/cli/cli_common.c
```diff
@@ -774,6 +774,94 @@ cli_set_mode(clixon_handle h,
...
+    }
+    if (cvec_len(argv) == 2){
+        runner = cv_string_get(cvec_i(argv, 0));
+        script_path = cv_string_get(cvec_i(argv, 1));
+    }
+    if (cvec_len(cvv) == 2){
+        script_path = cv_string_get(cvec_i(cvv, 1));
+    }
+
+    if ((pw = getpwuid(getuid())) == NULL){
```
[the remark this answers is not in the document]
khromenokroman: done

khromenokroman (commented):
done?

khromenokroman:
@olofhagsand @dima1308 work is done?

apps/cli/cli_common.c
```diff
@@ -774,6 +775,125 @@ cli_set_mode(clixon_handle h,
...
 
+/*!
+ * @brief Spawns a new process and runs the specified program within it
+ *
+ * The function creates a new child process and runs the specified program within it.
+ * Before launching the program in the child process, environment variables taken from cvv are set.
+ *
+ * The function checks the parameters passed to it. Situations where at least one function argument
+ * is missing, or when two arguments are present in both the function and the cvv vector are considered invalid.
+ *
```
olofhagsand, line 786: Please add a warning about safety here, essentially you could start any script from the command-line if you use some variants here. That may be OK, but a designer may also want to chroot after the fork to limit the space of accessible scripts.

olofhagsand (changes requested):
Just minor stuff, change and I will merge it.

olofhagsand:
Related: could one do something along the lines with: https://github.com/MontaVista-OpenSourceTechnology/clixon-backend-helper for CLI python scripts?

apps/cli/cli_common.c
```diff
@@ -774,6 +775,125 @@ cli_set_mode(clixon_handle h,
...
+ * Example usage:
+ * @code
+ *      run_program_err("Run program"), cli_start_program();
+ *      run_program_python3("Run program"), cli_start_program("python3");
+ *      run_program_python3_source_arg("Run program"), cli_start_program("python3", "/tmp/test.py");
+ *      run_program_python3_source_arg_vector("Run program") <source:rest>("Path program"), cli_start_program("python3");
+ *      run_program_python3_source_arg_vector_err("Run program") <source:rest>("Path program"), cli_start_program("python3", "/tmp/test2.py");
+ *      run_program_bash("Run program"), cli_start_program("bash");
```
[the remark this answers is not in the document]
khromenokroman: Снимок экрана от 2024-06-13 16-53-01

He works for me, but what does he write? Is there any mistake?

apps/cli/cli_common.c
```diff
@@ -774,6 +775,125 @@ cli_set_mode(clixon_handle h,
...
+    /* main run */
+    if ((pid = fork()) == 0) {
+        /* child process */
+        if ((work_dir ? chdir(work_dir) : chdir(pw.pw_dir)) < 0) {
+            clixon_err(OE_PLUGIN, errno, "chdir");
+        }
+        execlp(runner, runner, script_path, NULL);
+        clixon_err(OE_PLUGIN, errno, "Error run script");
+        return -1;
+    }else if(pid == -1){
```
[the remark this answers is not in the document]
khromenokroman: sorry, it is my mistake. done

apps/cli/cli_common.c
```diff
@@ -774,6 +775,125 @@ cli_set_mode(clixon_handle h,
...
+        clixon_err(OE_PLUGIN, errno, "fork");
+    }else{
```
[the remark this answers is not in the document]
khromenokroman: sorry, it is my mistake. done

apps/cli/cli_common.c
```diff
@@ -774,6 +775,125 @@ cli_set_mode(clixon_handle h,
...
+        /* parent process */
+        int status;
+        if (waitpid(pid, &status, 0) != pid ){
+            clixon_err(OE_PLUGIN, errno, "waitpid error");
+            goto done;
+        } else {
```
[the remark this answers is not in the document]
khromenokroman: sorry, it is my mistake. done

example/main/example_cli.cli
```diff
@@ -71,6 +71,7 @@ debug("Debugging parts of the system"){
 shell("System command"), cli_start_shell("bash");{
   <source:rest>("Single shell command"), cli_start_shell("bash");
 }
+
```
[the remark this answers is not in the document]
khromenokroman: it seemed to me that the separation between the teams was better visible, I fixed it :)

apps/cli/cli_common.c
khromenokroman: done

khromenokroman:
> Related: could one do something along the lines with: https://github.com/MontaVista-OpenSourceTechnology/clixon-backend-helper for CLI python scripts?

I'm not as well versed in the project as you are, as well as python and C to give an answer, but I'll read about it :)

Issue #292, opened 2021-12-16
Embedded Python API for callbacks


olofhagsand:
Clixon provides a C-API for callbacks.
A recurring topic in clixon is a foreign function interface enabling a developer to use another programming language when interacting with other tools. The reason is to increase usability, development speed, flexibility, etc.
So far, this has been up to the application developer, but there are multiple users having activities in this area.
It could be beneficial to maintain an optional ffi in Clixon, or maybe as an add-on to the core clixon functionality.
A quick slack investigation seem to be in favor of an embedded python C-API, although other alternatives (or even multiple) are not ruled out. Other alternatives could be golang, php, etc, but python seems the best option at this time.

akaevgen:
Hi Olof, 

Has there been any progress on this request? 🙂

rcmcdonald91:
This was something I originally proposed with Olof. It continues to be a side-project of mine. The idea here is to provide an interface that many different interpreters can plug into (Python, PHP, Lua, etc.). If you want to start a public discussion (design objectives, wishlists, etc), I say lets do it!

rcmcdonald91:
That's cool. Any reason not to instead build this as CPython module that wraps the Clixon client API?

dima1308:
Hi @olofhagsand maybe I'm missing something, but it is not clear to me why not use 3rd party Netconf library to communicate with CLIXON?
I did it in the past using https://github.com/ncclient/ncclient

olofhagsand:
Yes agree, a 3rd party client-side like ncclient is a prefered choice for a netconf client lib.
In the controller work we started out with a a python wrapper of the plugin API but instead made a more loosely coupled architecture where python code just uses netconf. I guess we could have used a 3rd party for the client (@krihal ?) but then the emphasis of pyapi is in the network service script code where a class structure is added.
The reason why we did not pursue a plugin wrapper was it was that for this application (the controller) it seemed an easier way forward to use a regular client.n
An embedded python API wrapper serves a different purpose and would be useful in other scenarions.
The

dima1308:
Nclient is a de-facto standard library for client side implementation in Python.
Developmant of another solution will take a significant effort.
From my point of view, if I can reuse existing solution that well maintained, I will do it.

cminyard:
Any word on this?  I am just starting with clixon, and this is an important piece we need for our application.

I have also done this on a couple of different libraries with a number of different languages.

[...]

PR #525, opened 2024-05-30
[RFC] Backend minor rework and improvement

cminyard:
I've been studying the backend and I found a few things.

The first two patches fix a bug and remove unnecessary code.

The next patch applies the DRY principle and consolidates some code.

The last two patches give an alternate interface to clixon_pseudo_plugin() that results in less code being used to call it, and replaces all the calls.  It also lets you add plugins from a loadable module and automatically cleans up any added plugins on a failure.

I've been looking at this because I have been looking at using clixon.  One thing I would like is the ability to have per-namespace backend plugins.  I think this makes sense (though I'm not 100% sure); this way we could have individual plugins  for each namespace without each plugin having to scan the entire XML tree for what they need.

It's also not completely clear to me  how to use the add/remove/change part of a transaction.  You are only getting a little bit of data.  You can search up the parents to find where you are in the tree, but it seems inefficient to have to do this on every plugin.  But that seems to be the only way you could use this.  Again, being able to do this on a per-namespace basis would seem more efficient.

### DIFF
--- a/apps/backend/backend_commit.c
+++ b/apps/backend/backend_commit.c
@@ -464,6 +464,177 @@ startup_commit(clixon_handle  h,
     goto done;
 }
 
+/*! Fetch the nsxml attribute from an element
+ *
+ * @param[in]  x       The xml element
+ * @retval     !NULL   Ok
+ * @retval     NULL    The attribute wasn't present
+ */
+static char *
+xml_nsxml_fetch(cxobj *x)
+{
+    cxobj *c = NULL;
+
+    while ((c = xml_child_each_attr(x, c)) != NULL) {
+        if (strcmp(xml_name(c), "xmlns") == 0)
+            return xml_value(c);
+    }
+    return NULL;
+}
+
+/*! Add a namespace entry to the transaction data's namespace info
+ *
+ * @param[in]  nss     The namespace info to add the transaction data to
+ * @param[in]  td      The transaction data to add to the namespace data
+ * @param[in]  ns      The namespace to add the info for
+ * @retval     !NULL   Ok
+ * @retval     NULL    The attribute wasn't present
+ */
+static int
+nss_add_ns(cvec *nss, transaction_data_t *td, const char *ns)
+{
+    cg_var *var;
+    cvec *xnvec;
+
+    var = cvec_find_var(nss, ns);
+    if (!var) {
+        xnvec = cvec_new(0);
+        if (!xnvec)
+            return -1;
+        var = cvec_add(nss, CGV_VOID);
+        if (!var) {
+            cvec_free(xnvec);
+            return -1;
+        }
+        if (!cv_name_set(var, ns)) {
+            cvec_free(xnvec);
+            return -1;
+        }
+        cv_void_set(var, xnvec);
+    } else {
+        xnvec = cv_void_get(var);
+    }
+    var = cvec_add(xnvec, CGV_VOID);
+    if (!var)
+        return -1;
+    cv_void_set(var, td);
+
+    return 0;
+}
+
+/*! Create sub-transaction for all changed transaction with an active namespace
+ *
+ * @param[in]  h       The clixon handle
+ * @param[in]  td      The transaction data to scan and add the sub-transactions to
+ * @retval     0       Ok
+ * @retval     -1      The attribute wasn't present
+ *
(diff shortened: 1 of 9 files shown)


olofhagsand:
> I've been looking at this because I have been looking at using clixon. One thing I would like is the ability to have per-namespace backend plugins. I think this makes sense (though I'm not 100% sure); this way we could have individual plugins for each namespace without each plugin having to scan the entire XML tree for what they need.

Initially we tried many methods, some with higher abstractions, but settled with the current API on a low abstraction level with few assumptions on how the programmer organizes the plugins vs YANG modules. It could be nice to have a more abstract interface based on modules/namespaces which is more coupled to the underlying YANG modules, but I would leave the current API in place for users that have other perspectives.
As an example of other perspectives, you could see a plugin based on a target system capability, (some OS configuration) that takes info from multiple YANG modules (some interface info, some DNS, maybe some ACL), and thus needs to collect info from multiple modules.
Historically, this is what lead us to the current (low-level) API.

> It's also not completely clear to me how to use the add/remove/change part of a transaction. You are only getting a little bit of data. You can search up the parents to find where you are in the tree, but it seems inefficient to have to do this on every plugin. But that seems to be the only way you could use this. Again, being able to do this on a per-namespace basis would seem more efficient.

You can also start from the top tree and traverse the tree in your modules by looking for changed flags as described here: https://clixon-docs.readthedocs.io/en/latest/backend.html#flags. But the callback need to keep track of which top-level symbols it has. Some users favor this model.

But I am open to adding a more high-level API more coupled to modules. I would assume one would register a plugin by modules/namespaces and then get localized info about what changed in that module. Maybe even with pointer to where in the YANG the change occurred.

cminyard:
Ok, I have pushed up some patches that add something like what we have talked about.  I agree with you, being able to see the whole tree would have and advantage for many applications, you wouldn't want to lose that capability.  This change won't work if the changes between namespaces are not independent.

Basically, after xml_diff() is called, it scans the source and target trees and detects changed top-level elements that have a namespace that a module has registered against.  It creates a new transaction_data item for those that have just those elements and stores it in the original transaction data.  The code assumes that the source and target trees are in the same order except for top-level added and deleted elements.

When calling the plugin callbacks, it will see if the plugin has a namespace.  If it does, it looks in the transaction data for transaction data items matching the namespace and calls the plugin callback for those.

I tried to marry simplicity and efficiency in this; this is the best I could come up with without reworking things a lot.

Some comments...

The namespace is set with clixon_add_plugin().  It's not in the api structure.  You could add it to the api structure and it would simplify the interface a bit and be more natural, but that would change the structure, and you couldn't re-use the same api structure for different namespaces.

This is not extensively tested, I have done a basic test of the new function and run the test suite.

This change consolidates a lot of common code in the backend plugin callbacks.  That's probably a good idea even without this change.

cminyard:
I fixed the issue that showed up.  I thought it was a base issue, but how all the tests work is a little confusing.  Anyway, it's fixed.

There is an issue with test_datastore_multi.sh, but I was able to reproduce that without my patches, and it only happens occasionally.  I get:
```
Test 35(35) [Check running subfile not changed]

Error in Test35 [Check running subfile not changed]:
Expected
Timestamp not changed

Received: 1717430952 != 1717430953

Diff between Expected and Received:
1,3c1,3
< 0000000   1   7   1   7   4   3   0   9   5   2       !   =       1   7
< 0000020   1   7   4   3   0   9   5   3  \n
< 0000031
---
> 0000000   T   i   m   e   s   t   a   m   p       n   o   t       c   h
> 0000020   a   n   g   e   d  \n
> 0000026
```

olofhagsand:
?> There is an issue with test_datastore_multi.sh, but I was able to reproduce that without my patches, and it only happens occasionally. I get:

Yes, it is a known that happens now and then

cminyard:
I reworked the patches a bit.

The first four patches are ones I consider to be "good ideas in general".  They consolidate redundant code and things like that.  One fixes what I think is a bug.

Per the discussions in https://github.com/clicon/clixon/issues/292, I'm beginning to think it is better to not add a namespace-based plugin to the main clixon library, since API changes will be required to make it work right.  I can reuse a lot of the code written here to do that.

cminyard:
Doing this as an external plugin.  No need for most of these changes.  I'll try to open another request with the generic improvements.

PR #565, opened 2024-10-15
Allow disabling of netconf notification

mpereh:


olofhagsand (changes requested):
I agree it could be useful to have a knob for turning off netconf notifications.
But I am interested in the usecase, in which way did it fail / why was it necessary?
But there are other ways to design the knob, and this is knob is not right: CLICON_NETCONF_NOTIFICATION_CAP is not defined in clixon-config.yang.
If it is not defined there (and added to the config file) then this will(should) result in a sanity check error when loading the config.
There are also other ways, such as compile time using clixon_custom.h, or make a --enable / --with rule in configure.ac.
One could also define it as a FEATURE possibly.

olofhagsand:
@mpereh is this still an active PR?

mpereh:
@olofhagsand not anymore, we found a way around

PR #569, opened 2024-10-21
Added build_deb.sh and version.sh

krihal:


olofhagsand (commented):
This version.sh is already under scripts/
Consider adding build_deb.sh there instead.
Maybe we should move those in cligen/ too to a scripts/ dir?

PR #584, opened 2024-12-06
replace select to poll

khromenokroman:
@olofhagsand Faced with such a problem, if there are a lot of descriptors, then select cannot be used since it has a limit of 1024


rcmcdonald91:
> @olofhagsand Faced with such a problem, if there are a lot of descriptors, then select cannot be used since it has a limit of 1024

I wonder if we should instead adopt an async i/o library like libevent, libuv, etc. so each target platform gets the best non-blocking i/o interface available (e.g. epoll on Linux, kqueue on *BSD, etc.) for "free."

khromenokroman:
> > @olofhagsand Faced with such a problem, if there are a lot of descriptors, then select cannot be used since it has a limit of 1024
> 
> I wonder if we should instead adopt an async i/o library like libevent, libuv, etc. so each target platform gets the best non-blocking i/o interface available (e.g. epoll on Linux, kqueue on *BSD, etc.) for "free."

better then io_uring then :)

olofhagsand:
OK, the controller especially opens many file descriptors, but I dont think clixon itself (for end-systems) has this need.
Though, the patch ^ is for clixon_event_poll() which is used only occasionally
But the main loop using select is in clixon_event_loop(), which is the main work. Just changing in one and not the other does not make sense.
Personally, I prefer the simplicity of poll /select over libevent and friends, and I think many in the embedded field do as well.

khromenokroman:
> OK, the controller especially opens many file descriptors, but I dont think clixon itself (for end-systems) has this need. Though, the patch ^ is for clixon_event_poll() which is used only occasionally But the main loop using select is in clixon_event_loop(), which is the main work. Just changing in one and not the other does not make sense. Personally, I prefer the simplicity of poll /select over libevent and friends, and I think many in the embedded field do as well.

I also want to replace clixon_event_poll with poll later because I don't have enough select size of 1024. I didn't want to make a big pull request, so I decided to split it into parts

khromenokroman:
> OK, the controller especially opens many file descriptors, but I dont think clixon itself (for end-systems) has this need. Though, the patch ^ is for clixon_event_poll() which is used only occasionally But the main loop using select is in clixon_event_loop(), which is the main work. Just changing in one and not the other does not make sense. Personally, I prefer the simplicity of poll /select over libevent and friends, and I think many in the embedded field do as well.

there may also be such a problem in CLIOXN itself, for example, if there is a DPDK and there are hugepages = 2048 and we need to allocate 2GB or more, it turns out that the descriptors will be busy and the size of 1024 will not be enough

khromenokroman:
I had this information only once, I allocated 2 GB of huge pages and because of this, the cli did not work correctly

I executed `gdb -p $(pid clixon_backend) -ex bt`

and I saw that there the descriptor was 1028 and the nfds was 1024

khromenokroman:
I rewrote the standard functions to poll and everything worked correctly
```
#define MAX_EVENTS 2048
/*!
 * Poll to see if there is any data available on this file descriptor.
 *
 * @param[in]  fd   File descriptor
 * @retval     1    Something to read on fd
 * @retval     0    Nothing to read/empty fd
 * @retval    -1    Error
 */
int clixon_event_poll(int fd) {
    struct pollfd pfd;
    int retval;

    // Настройка структуры pollfd
    pfd.fd = fd;
    pfd.events = POLLIN; // Интересуемся событиями чтения

    // Выполняем poll с ожиданием 0 миллисекунд, чтобы сразу проверить состояние
    retval = poll(&pfd, 1, 0);

    if (retval < 0) {
        clixon_err(OE_EVENTS, errno, "poll");
    }

    // Если произошла ошибка, poll вернет <0; если окончание времени ожидания вернет 0;
    // если дескриптор готов к чтению, вернёт >0.
    return retval;
}

/*! Dispatch file descriptor events (and timeouts) by invoking callbacks.
 *
 * @param[in] h  Clixon handle
 * @retval    0  OK
 * @retval   -1  Error: eg select, callback, timer, 
 * @note There is an issue with fairness between timeouts and events
 *       Currently a socket that is not read/emptied properly starve timeouts.
 *       One could try to poll the file descriptors after a timeout?
 */
int
clixon_event_loop(clixon_handle h)
{
    struct event_data *e;
    int retval = -1;
    struct pollfd fds[MAX_EVENTS];
    int nfds = 0;

    while (clixon_exit_get() != 1) {
        nfds = 0;
        for (e = ee; e; e = e->e_next) {
            if (e->e_type == EVENT_FD) {
                fds[nfds].fd = e->e_fd;
                fds[nfds].events = POLLIN;  // флаги, которые нужно следить
                nfds++;
            }
        }

        int timeout = -1;  // Бесконечное ожидание
        if (ee_timers != NULL) {
            struct timeval t0, t;
            gettimeofday(&t0, NULL);
            timersub(&ee_timers->e_time, &t0, &t);
            timeout = t.tv_sec * 1000 + t.tv_usec / 1000; // Преобразование в миллисекунды
        }

        int n = poll(fds, nfds, timeout);

        if (n == -1) {
            if (errno == EINTR) {
                continue;
            }
            clixon_err(OE_EVENTS, errno, "poll");
            goto err;
        }

        if (n == 0) {
            // Таймаут
            e = ee_timers;
            ee_timers = ee_timers->e_next;
            if ((*e->e_fn)(0, e->e_arg) < 0) {
                free(e);
                goto err;
            }
            free(e);
            continue;
        }

        for (int i = 0; i < nfds; i++) {
            if (fds[i].revents & POLLIN) {
                // Обработка события
                for (e = ee; e; e = e->e_next) {
                    if (e->e_type == EVENT_FD && e->e_fd == fds[i].fd) {
                        if ((*e->e_fn)(e->e_fd, e->e_arg) < 0) {
                            goto err;
                        }
                        break;
                    }
                }
            }
        }
    }

    retval = 0;

err:
    return retval;
}
```

khromenokroman:
but I have a question with the main function, there is something to optimize, but I think it's better to do it after we change this small function so as not to make a big pull request

olofhagsand:
OK, so you are saying with hugepages, the file descriptor numbers start at 1024, it is not the number of file descriptors that exceed 1000.

lib/src/clixon_event.c
```diff
@@ -346,24 +347,27 @@ clixon_event_unreg_timeout(int (*fn)(int, void*),
...
-{
-    int            retval = -1;
-    fd_set         fdset;
-    struct timeval tnull = {0,};
-
-    FD_ZERO(&fdset);
-    FD_SET(fd, &fdset);
-    if ((retval = select(FD_SETSIZE, &fdset, NULL, NULL, &tnull)) < 0)
-        clixon_err(OE_EVENTS, errno, "select");
+int clixon_event_poll(int fd) {
```
olofhagsand, line 358: follow style guidelines:
int
clixon_event_poll

lib/src/clixon_event.c
```diff
@@ -346,24 +347,27 @@ clixon_event_unreg_timeout(int (*fn)(int, void*),
     return found?0:-1;
 }
 
-/*! Poll to see if there is any data available on this file descriptor.
+/*!
```
olofhagsand, line 350: plz revert this LF

lib/src/clixon_event.c
```diff
@@ -53,6 +53,7 @@
 #include <sys/param.h>
 #include <sys/types.h>
 #include <sys/time.h>
+#include <poll.h>
```
olofhagsand, line 56: Can you please add a compile-time constant around the poll change, so we can introduce it gradually, and have it as experimental code for a while
I would suggest :
`#ifdef CLIXON_EVENT_POLL`
and define the constant in clixon_custom.h

dima1308:
> OK, so you are saying with hugepages, the file descriptor numbers start at 1024, it is not the number of file descriptors that exceed 1000.

@olofhagsand  file descriptor numbers can be greater than 1024, for example, if the process opens many files for any purpose that is not directly related to the CLIXON usage.

lib/src/clixon_event.c
khromenokroman: > and define the constant in clixon_custom.h

@olofhagsand I didn't quite understand why clixon_custom.h is needed.

khromenokroman: > Can you please add a compile-time constant around the poll change, so we can introduce it gradually, and have it as experimental code for a while
> I would suggest :
> `#ifdef CLIXON_EVENT_POLL`

@olofhagsand I did this `./configure --enable-event-poll`

lib/src/clixon_event.c
khromenokroman: > plz revert this LF

@olofhagsand done

lib/src/clixon_event.c
khromenokroman: > follow style guidelines: int clixon_event_poll

@olofhagsand yes, the auto-formatting worked, it's done

lib/src/clixon_event.c
khromenokroman: > Can you please add a compile-time constant around the poll change, so we can introduce it gradually, and have it as experimental code for a while I would suggest : `#ifdef CLIXON_EVENT_POLL` and define the constant in clixon_custom.h

I think I got it, thanks :)

olofhagsand: > > and define the constant in clixon_custom.h
> 
> @olofhagsand I didn't quite understand why clixon_custom.h is needed.

It is not needed when you added it in configure.ac, so it can be removed

olofhagsand: > > Can you please add a compile-time constant around the poll change, so we can introduce it gradually, and have it as experimental code for a while
> > I would suggest :
> > `#ifdef CLIXON_EVENT_POLL`
> 
> @olofhagsand I did this `./configure --enable-event-poll`

When you add this to configure.ac, you also need to run "autoheader" and then add include/clixon_config.h.in.
Then you dont need the clixon_custom.h part.

olofhagsand:
As noted above:
1) You can remove clixon_custom.h since you added it (better) in configure.ac. Just run autoheader so clixon_config.h.in gets the new constant template.
2) I note the main select is not in this PR. Had you planned to add it in a follow-up PR?

khromenokroman:
> 1. You can remove clixon_custom.h since you added it (better) in configure.ac. Just run autoheader so clixon_config.h.in gets the new constant template.

@olofhagsand  done

khromenokroman:
> 2\. I note the main select is not in this PR. Had you planned to add it in a follow-up PR?

@olofhagsand yes, there will be another pull request in which I will still change the main function, I did not want to make a big pull

PR #603, opened 2025-04-15
Optimize ptr2ptr mapping

ivan4th:
Use hash table instead of linear search.


ivan4th:
Made a PR with pointer hash tables: #608
Closing this PR.

PR #620, opened 2025-08-29
Allow module's namespace upgrade

askorichenko:
Allow upgrade of module's namespace 
```
From:
    <module>
      <name>modname</name>
      <revision>rev-prev</revision>
      <namespace>ns-prev</namespace>
    </module>
To:
    <module>
      <name>modname</name>
      <revision>rev-new</revision>
      <namespace>ns-new</namespace>
    </module>

Via msdiff node with old revision and new namespace:
    <module>
      <name>modname</name>
      <revision>rev-prev</revision>
      <namespace>ns-new</namespace>
    </module>
```


olofhagsand:
There are some upgrade tests, see test/test_upgrade*.sh. 
Would it be possible to add a regression test for this case?

PR #634, opened 2025-11-28
Better handle size_t on 32bit system for cli_show

Ansuel:
Size_t have different size on 32bit system and it's normally 4 bytes instead of 8 bytes.

This cause compilation warning and also error for parse_uint64.

To better handle this, introduce a new variable to handle case where size_t can be used as is and change sz type to uint64_t to follow parse_uint64 requirement.

This also handle corner case where value might be cut after parsing a value of 4 bytes expected to be 8 bytes.

While at it also change the cligen_output to %zu where size_t type is used to further mute additional compilation warning.

### DIFF
--- a/apps/cli/cli_show.c
+++ b/apps/cli/cli_show.c
@@ -2012,7 +2012,8 @@ cli_show_statistics(clixon_handle h,
     uint64_t    nr;
     uint64_t    tnr0;
     uint64_t    tnr = 0;
-    size_t      sz;
+    uint64_t    sz;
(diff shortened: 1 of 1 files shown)


Ansuel:
Thanks for enabling CI. Will solve the new error for 64bit I assume?

Il Dom 30 Nov 2025, 10:05 Olof Hagsand ***@***.***> ha
scritto:

olofhagsand:
Yes, will fix

Issue #635, opened 2025-12-06
when on uses leaks to other modules via shared grouping


yamontv:
## Summary
When two modules import a shared grouping, a `when` clause applied to a `uses` of that grouping in one module is incorrectly enforced on other `uses` of the same grouping in other modules. This breaks OpenConfig QoS + ACL combinations: simply importing `openconfig-acl` causes edits under `openconfig-qos` classifiers to fail with an ACL `when` guard.

## Minimal self‑contained repro

Place these four modules on the YANG path and set `test-platform` as the main module.

**test-base.yang** (shared grouping):
```yang
module test-base {
  yang-version 1.1;
  namespace "urn:example:test-base";
  prefix tb;
  grouping ipv4-fields {
    container ipv4 {
      leaf dscp { type uint8; }
    }
  }
}
```

**test-acl.yang** (uses grouping with `when`):
```yang
module test-acl {
  yang-version 1.1;
  namespace "urn:example:test-acl";
  prefix ta;
  import test-base { prefix tb; }
  container acl {
    leaf type {
      type enumeration { enum ACL_IPV4; enum ACL_IPV6; }
    }
    container conditions {
      uses tb:ipv4-fields {
        when "type = 'ACL_IPV4'";
      }
    }
  }
}
```

**test-qos.yang** (uses grouping without `when`):
```yang
module test-qos {
  yang-version 1.1;
  namespace "urn:example:test-qos";
  prefix tq;
  import test-base { prefix tb; }
  container classifiers {
    list classifier {
      key "name";
      leaf name { type string; }
      container conditions {
        uses tb:ipv4-fields;
      }
    }
  }
}
```

**test-platform.yang** (entry point):
```yang
module test-platform {
  yang-version 1.1;
  namespace "urn:example:test-platform";
  prefix tp;
  import test-acl { prefix ta; }
  import test-qos { prefix tq; }
}
```

### Steps to reproduce
1) Load `test-platform` as main YANG.
2) In CLI:
```
set classifiers classifier C1 conditions ipv4 dscp 46
```

**Expected:** Succeeds; `test-qos` has no `when` on `conditions/ipv4`.

**Actual:** Validation fails with:
```
[...]

PR #636, opened 2025-12-07
Yang/when: store guards on copies only and drop origin mirroring

yamontv:
- Keep when mappings strictly per-node; remove origin fallback and
  mirroring in get/set.
- Let ys_cp_one propagate existing when guards during duplication
  so grouping instances inherit them without ys_orig.

https://github.com/clicon/clixon/issues/635


yamontv:
Brilliant suggestion - thanks for pointing out such a simple solution! I’ve tested it; everything works flawlessly.

Issue #637, opened 2025-12-07
Clixon ignores imports declared in submodules


yamontv:
## Summary

When a main YANG module relies on imports declared only in its included submodules, Clixon’s CLI/leafref resolution fails with “Prefix \<prefix\> does not have an associated namespace”. The namespace context is built from the top-level module imports only, so prefixes imported in submodules are missing. This breaks OpenConfig QoS because `openconfig-qos`’s `interface-ref` lives in the submodule `openconfig-qos-interfaces`, which imports `openconfig-interfaces` (prefix `oc-if`). The CLI cannot complete or evaluate `interface-ref` and aborts with an “oc-if” namespace error.

## Minimal reproducer

Use these four files:

`test-platform.yang` (entry point)

```yang
module test-platform {
  yang-version 1.1;
  namespace "urn:test:platform";
  prefix tp;

  import test-main { prefix tm; }
  import test-if   { prefix tif; }

  /* Entry point that pulls in the test modules */
}
```

`test-main.yang`

```yang
module test-main {
  yang-version 1.1;
  namespace "urn:test:main";
  prefix tm;

  include test-sub;

  container top {
    uses sub-top;
  }
}
```

`test-sub.yang`

```yang
submodule test-sub {
  belongs-to test-main { prefix tm; }
  yang-version 1.1;

  import test-if { prefix tif; }

  grouping sub-top {
    container iface-ref {
      leaf interface {
        type leafref {
          path "/tif:interfaces/tif:interface/tif:name";
        }
      }
    }
  }
}
```

`test-if.yang`

```yang
module test-if {
  yang-version 1.1;
  namespace "urn:test:if";
  prefix tif;

  container interfaces {
    list interface {
      key "name";
      leaf name {
        type string;
      }
    }
  }
}
```

### Steps

1) Load test-platform into Clixon (e.g., via `CLICON_YANG_MODULE_MAIN`).
2) Enter CLI and try to configure the leaf:

   ```cli
   set top iface-ref interface <TAB>
   ```

3) Observe the error:

   ```cli
   xpath_append: 160: Database error: Prefix tif does not have an associated namespace
   CLI command error
   ```

### Expected

Tab completion/leafref resolution should work; `tif` should be known via the submodule import.

### Actual

Clixon reports `Prefix tif does not have an associated namespace` because the submodule imports are not added to the namespace context used by the CLI/XPath/leafref code.

## Real-world impact (OpenConfig)

[...]

PR #638, opened 2025-12-07
Resolve imports in lexical scope and fix leafref/xpath resolution

yamontv:
- Add yang_imports_foreach_scope() and use it to resolve imports
  bottom-up from the lexical origin (ys_orig) through the
  belongs-to chain, skipping sibling submodules.
- Make nsctx building use the instantiated node’s own prefix/namespace,
  but gather import prefixes from the lexical scope;
  keep xml_nsctx_yangspec to module defaults only to avoid cross-module
  prefix collisions.
- Scope prefix/namespace lookups (yang_find_module_by_prefix,
  yang_find_prefix_by_namespace) to the lexical origin, with own-module
  matching via the instantiated node; accept submodules in XPath
  absolute evaluation.
- Leafref validation now resolves paths from the lexical origin, falls
  back to the instantiated tree if needed, and guards against missing
  cv templates to avoid crashes.

Fixes grouping/augment leafref resolution and prevents wrong-prefix
matches when submodules reuse prefixes differently.
https://github.com/clicon/clixon/issues/637

### DIFF
--- a/lib/clixon/clixon_yang_module.h
+++ b/lib/clixon/clixon_yang_module.h
@@ -81,5 +81,8 @@ yang_stmt *yang_find_module_by_name(yang_stmt *yspec, const char *name);
 int        yang_metadata_annotation_check(cxobj *x, yang_stmt *ymod, int *ismeta);
 int        yang_metadata_init(clixon_handle h);
 int        yang_lib2yspec(clixon_handle h, cxobj *yanglib, const char *mntpnt, const char *domain, yang_stmt *yspec);
+int        yang_imports_foreach_scope(yang_stmt *ys, yang_stmt *yspec,
+                                      int (*cb)(yang_stmt *yimport, void *arg),
+                                      void *arg);
 
 #endif  /* _CLIXON_YANG_MODULE_H_ */
--- a/lib/src/clixon_xml_nsctx.c
+++ b/lib/src/clixon_xml_nsctx.c
@@ -73,6 +73,38 @@
 #include "clixon_xml_sort.h"
 #include "clixon_xml_nsctx.h"
 
+/* Context for adding imports to a namespace context */
+struct nsctx_import_ctx {
+    cvec *nc;
+    yang_stmt *yspec;
+};
+
+/* Callback used by yang_imports_foreach_scope to add import prefix/ns to nsctx */
+static int
+nsctx_add_import_cb(yang_stmt *yimport, void *arg)
+{
+    struct nsctx_import_ctx *c = arg;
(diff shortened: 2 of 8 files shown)


olofhagsand:
Thanks for detecting this and for your submission.
The code solves the openconfig issue given in #637 .
However, I dont think it solves a more general case, which would be nice to solve while this is addressed.
But it is a good start, the code is good quality.
The issue/testcase has the following hierarchy:
```
module test-main
  +-- include sub-module test-sub
        +-- import test-if { prefix tif; }
```
In the commit, yang_imports_foreach_module() traverses the modules from 'test-main' to 'test-if' and finds a match top-down.
However, assume a slightly more elaborate hierarchy, such as:
```
module test-main
  +-- include sub-module test-sub
  |   +-- include sub-module test-subsub
  +-- include sub-module test-sub2
``` 
with different import/prefix mappings in test-sub and test-sub2, then yang_imports_foreach_module() would:
1. choose a first map that might be wrong (if test-sub and test-sub2 have same import with different prefix)
2. not traverse to test-subsub (I see no recursion in yang_imports_foreach_module())?

Therefore, I think a correct resolver needs to start from where the yang node (specifically in a grouping) is defined and check upwards in the hierarchy rather than top-down. 
See rfc7950 sec 7.12: 
```
  Prefix mappings [...] are evaluated in the hierarchy where the "grouping" statement appears.  
```
For example, consider the test-case with the following structure:
```
module test-main
  +-- include sub-module test-sub
  |      +-- import test-if { prefix tif; }
  |      +--   grouping sub-top {
  |               container iface-ref {
  |                leaf interface {  // ys->ys_orig
  |                  type leafref {
  |                    path "/tif:interfaces/tif:interface/tif:name";
  (+-- uses sub-top)
  +-- container iface-ref {
          leaf interface { // ys
            type leafref {
              path "/tif:interfaces/tif:interface/tif:name";
```
The failing call to `yang_find_module_by_prefix(ys, "tif")` is in the solution at hand called to the module test-main, and then iterates eventually to sub-module test-sub and the import matching is found.
I think the correct method is to instead follow ys->ys_orig, if defined, (ie to the original grouping definition) and make the  prefix resolving there.
ys_orig is only defined for grouping and augments and would in that case not need to traverse sibling sub-modules. 
I think however one need to make traversal code upwards in the hierarchy if the prefix is not found in the sub-module.
And also for the corresponding cases for `yang_find_prefix_by_namespace()`,  `xml_nsctx_yang()` and `xml_nsctx_yangspec()`
What do you think?

PR #640, opened 2026-01-04
Preserve OpenConfig layout in test images

yamontv:
The stage‑2 COPY --from=0 /usr/local/share/openconfig/* flattened the public directory into /usr/local/share/openconfig, so the expected /usr/local/share/openconfig/public path was missing at runtime. This change copies the whole directory tree so
OPENCONFIG=/usr/local/share/openconfig/public works again in the test container.

### DIFF
--- a/docker/test/Dockerfile
+++ b/docker/test/Dockerfile
@@ -144,7 +144,7 @@ RUN adduser -D -H clicon
 
 COPY --from=0 /clixon/build/ /
 COPY --from=0 /usr/local/share/yang/ /usr/local/share/yang/
-COPY --from=0 /usr/local/share/openconfig/* /usr/local/share/openconfig/
+COPY --from=0 /usr/local/share/openconfig/ /usr/local/share/openconfig/
 COPY --from=0 /usr/local/share/mib-yangs/* /usr/local/share/mib-yangs/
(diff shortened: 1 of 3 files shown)


yamontv:
This is what I have without my fix:
```console
/ # cd /usr/local/share/openconfig/
/usr/local/share/openconfig # tree
.
├── CONTRIBUTING.md
├── LICENSE
├── README.md
├── cloudbuild.yaml
├── doc
│   ├── Integrated-Circuit_pipeline_aggregated_counters_guide.md
│   ├── component_fault_overview.md
```
It is a known Docker issue (or feature) https://stackoverflow.com/questions/30215830/dockerfile-copy-keep-subdirectory-structure
https://github.com/moby/moby/issues/15858#issuecomment-136552399

With Docker’s documented behavior, our glob with * is effectively “copy each match into the destination directory”, and when a match is a directory, Docker copies that directory’s contents into the destination (merging). So if * matches only public, you’ll see the contents of public land directly under /usr/local/share/openconfig/ - which looks like "flattening”.

https://docs.docker.com/reference/dockerfile/#copy

> If the source is a directory, the contents of the directory are copied, including filesystem metadata. The directory itself isn't copied, only its contents.

I hope you enabled DOCKERFLAGS="--no-cache" in the docker/test/Makefile

PR #642, opened 2026-01-08
Merge master branch to relax-validate

adrikacang:
Previously the relax-validate branch was failed to build, issue solved when merging to master branch


olofhagsand:
Configure option: add it in configure.ac
Runtime: Add a new option in clixon-config.yang
However, the whole option is a catch-all for a specific device that does not comply to netconf/yang rules. One should probably spend more time in identifying the compliance and try to fix them individually. Enabling this option will most probably add troubles further down the road.

adrikacang:
@olofhagsand Yes, I would need this branch for development purpose. Thank you.

PR #651, opened 2026-01-27
Fix identityref prefix namespaces during backend edit-config

yamontv:
Resolve missing identityref prefixes in edit-config by looking up the YANG module namespace on the backend and injecting the xmlns binding before validation.


olofhagsand:
Is this PR obsoleted by https://github.com/clicon/clixon/pull/649 ? If so please close it

PR #661, opened 2026-03-17
fix: yang2cli crash

bhumong:
## Summary

  This PR brings targeted fixes from epic/auto-cli-grouping to improve AutoCLI/grouping behavior and
  grouping resolution across module/submodule contexts, plus a regression test.

  ## Problem

  AutoCLI/grouping flows had edge cases where grouping resolution and context handling could fail or
  become brittle:

  - grouping lookup could fail depending on statement origin context;
  - when evaluation could miss raw namespace-context fallback behavior;
  - error messages for unresolved grouping were not informative enough for debugging.

  ## Changes

  ### 1. Improve grouping resolution fallback

  - File: lib/src/clixon_yang_parse_lib.c
  - ys_grouping_resolve() now falls back to resolving from yang_orig_get(yuses) when direct lexical
    resolution returns no grouping.
  - This helps cases where grouped uses are evaluated via copied/origin-linked statements.

  ### 2. Improve AutoCLI unresolved-grouping diagnostics

  - File: lib/src/clixon_autocli_generate.c
  - Error message now reports module context (grouping <name> not found in <module>) instead of a blank
    suffix.
  - Makes troubleshooting unresolved grouping references much clearer.

  ### 3. Add fallback evaluation for when in xml mapping

  - File: lib/src/clixon_xml_map.c
  - In yang_check_when_xpath(), when canonical evaluation returns false, code now attempts evaluation
    using raw XPath + raw namespace context from yang_when_xpath_get(...).
  - Improves robustness for namespace/prefix-sensitive when expressions.

  ### 4. Minor generated-source formatting delta

  - File: apps/cli/cli_generate.c
  - No functional logic change (newline-at-EOF delta).

  ## Test Added

  ### New regression test

  - File: test/test_autocli_grouping_origin.sh

  What it validates:

  - Grouping defined in submodule and consumed through another module is resolved correctly.
  - Nested uses inside grouping works with AutoCLI generation.
  - CLI operations (set, validate, commit, show config) succeed for the generated tree.
  - Test expectation is resilient to pretty-printed XML output formatting.

  ## Scope

  - Includes only AutoCLI/CLI/grouping-related files:
      - apps/cli/cli_generate.c
      - lib/src/clixon_autocli_generate.c
      - lib/src/clixon_xml_map.c
      - lib/src/clixon_yang_parse_lib.c
      - test/test_autocli_grouping_origin.sh


apps/cli/cli_generate.c
```diff
@@ -740,4 +740,4 @@ yang2cli_init(clixon_handle h)
     retval = 0;
  done:
     return retval;
-}
+}
```
olofhagsand, line 743: This seems wrong, file should end with a LF (typo?)

PR #666, opened 2026-03-26
Increased the size of the error message line

khromenokroman:
@olofhagsand I came across a situation where the 256 size is not always enough to output a message


olofhagsand:
Refactored the error buffer to a dynamic string instead, see:
https://github.com/clicon/clixon/commit/9b4033d8c2d36177da1268fe635d88fe40859820

PR #668, opened 2026-04-21
cli_show_common: replace rpc_translate_format with direct per-node formatting for JSON, NETCONF, XML, TEXT, and CLI output.

krihal:


olofhagsand:
No that reverts the CLI optimization that reduced the YANG in he CLI for mount-points.
What is needed here is bundle the get-config AND the formatting by the backend.

Issue #670, opened 2026-04-30
XML/YANG binding uses wrong sibling namespace for same local node names


bhumong:
## Summary

Clixon can bind/render XML against the wrong YANG module when sibling config
subtrees have the same local node names but different namespaces.

In the reproduced case, two Cisco XR AAA model families are present:

- classic/locald AAA:
  `http://cisco.com/ns/yang/Cisco-IOS-XR-aaa-locald-cfg`
- UM AAA:
  `http://cisco.com/ns/yang/Cisco-IOS-XR-um-aaa-task-user-cfg`

Both contain a similar path shape:

```text
aaa / usernames / username / secret
```

but the valid children under `secret` differ:

```xml
<!-- locald AAA -->
<secret>
  <type>type10</type>
  <secret10>xx</secret10>
</secret>

<!-- UM AAA -->
<secret>
  <ten>xx</ten>
</secret>
```

The UM XML is valid against the UM YANG module, but during show/translate it is
bound as if it belonged to the locald module. This causes:

```text
Failed to find YANG spec of XML node: ten with parent: secret
```

## Actual Behavior

After valid XML with both namespaces is loaded, `show configuration` fails:

```text
clixon_rpc_translate_format: Netconf error: Debug: application unknown-element
Failed to find YANG spec of XML node: ten with parent: secret
in namespace: http://cisco.com/ns/yang/Cisco-IOS-XR-um-aaa-task-user-cfg
in file: .../Cisco-IOS-XR-aaa-locald-cfg@2022-11-28.yang
<bad-element>ten</bad-element>
```

The key problem is that the error reports the UM namespace but the locald YANG
file. That indicates the binder selected the wrong sibling/module context.

## Expected Behavior

Clixon should bind each XML node using its own namespace and the correct YANG
module, even when a sibling subtree has the same local path names.

For the reproducer:

- `<secret><secret10>...</secret10></secret>` should bind to
  `Cisco-IOS-XR-aaa-locald-cfg`.
- `<secret><ten>...</ten></secret>` should bind to
  `Cisco-IOS-XR-um-aaa-task-user-cfg`.
- `show configuration` should render both subtrees successfully.

## Reproducer

A focused Clixon regression reproducer was added at:

[test_xml_bind_namespace_collision.sh](https://github.com/user-attachments/files/27234970/test_xml_bind_namespace_collision.sh)

```text
clixon/test/test_xml_bind_namespace_collision.sh
```

[...]

PR #672, opened 2026-05-05
fix(xml_bind): namespace collision

bhumong:
propose to fix issue [670](https://github.com/clicon/clixon/issues/670)

### DIFF
--- a/lib/src/clixon_xml_bind.c
+++ b/lib/src/clixon_xml_bind.c
@@ -167,13 +167,22 @@ populate_self_parent(clixon_handle h,
     yang_stmt *yspec1;
     yang_stmt *ymod;
     int        ret;
+    int        ns_resolved = 0;
 
     name = xml_name(xt);
     /* optimization for massive lists - use the first element as role model */
     if (xsibling &&
         xml_child_nr_type(xt, CX_ATTR) == 0){
-        y = xml_spec(xsibling);
-        goto set;
+        if ((y = xml_spec(xsibling)) != NULL){
+            if (xml2ns(xt, xml_prefix(xt), &ns) < 0)
+                goto done;
+            ns_resolved++;
+            nsy = yang_find_mynamespace(y);
+            if (nsy != NULL && ns != NULL && strcmp(ns, nsy) == 0)
+                goto set;
+            y = NULL;
+            nsy = NULL;
+        }
     }
     if ((xp = xml_parent(xt)) == NULL){
         if (xerr &&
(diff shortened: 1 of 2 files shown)


test/test_xml_bind_namespace_collision.sh
```diff
@@ -0,0 +1,206 @@
...
+  </aaa>
+</config>
+EOF
+
+new "test params: -f $cfg"
+if [ $BE -ne 0 ]; then
+    new "kill old backend"
+    $clixon_backend -z -f $cfg >/dev/null 2>&1 || true
+    new "start backend -s startup -f $cfg"
+    $clixon_backend -D $DBG -s startup -f $cfg
```
olofhagsand, line 176: Please use the start_backend to mimic other tests

test/test_xml_bind_namespace_collision.sh
```diff
@@ -0,0 +1,206 @@
...
+    "<usernames xmlns=\"http://cisco.com/ns/yang/Cisco-IOS-XR-aaa-locald-cfg\">" \
+    "<type>type10</type>" \
+    "<secret10>xx</secret10>" \
+    "<aaa xmlns=\"http://cisco.com/ns/yang/Cisco-IOS-XR-um-aaa-cfg\">" \
+    "<usernames xmlns=\"http://cisco.com/ns/yang/Cisco-IOS-XR-um-aaa-task-user-cfg\">" \
+    "<ten>xx</ten>"
+
+if [ $BE -ne 0 ]; then
+    new "Kill backend"
+    $clixon_backend -z -f $cfg
```
olofhagsand, line 197: Please use stop_backend

lib/src/clixon_xml_bind.c
```diff
@@ -168,13 +168,22 @@ populate_self_parent(clixon_handle h,
...
     /* optimization for massive lists - use the first element as role model */
     if (xsibling &&
         xml_child_nr_type(xt, CX_ATTR) == 0){
-        y = xml_spec(xsibling);
-        goto set;
+        if ((y = xml_spec(xsibling)) != NULL){
+            if (xml2ns(xt, xml_prefix(xt), &ns) < 0)
+                goto done;
+            ns_resolved++;
+            nsy = yang_find_mynamespace(y);
```
olofhagsand, line 181: NULL return is actually an error and should goto fail, but is harmless here, and if checked should have been tested earlier

olofhagsand (commented):
The xsibling code further down in xml_bind_yang0_opt:
```
 else if (xsibling &&
                 (xs = xml_find_type(xsibling, prefix, name, CX_ELMNT)) != NULL){
```
also suffers from a similar issue.
Summarizing, the fix is correct, but the code needs a restructuring for the xsibling optimization along with the yang check to be clearer.
I will pull this in and modify it in master.

PR #676, opened 2026-05-25
restconf: wait for complete HTTP header before parsing

hoxnox:
Large HTTP headers (e.g. JWT Bearer tokens >2KB) may be fragmented across multiple TCP segments. The current code calls clixon_http1_parse_string() on every read and, if parsing fails, uses SSL_pending() / clixon_event_poll() to decide whether more data is coming. This check is racy: when the next TCP segment hasn't yet arrived in the kernel buffer (common over WAN links with
>1ms inter-packet delay), poll returns 0 and a valid request is rejected
with 400 Bad Request / malformed-message.

Fix this by checking for the HTTP header terminator (\r\n\r\n) before attempting to parse. If the terminator is not present, return to the event loop; the next readable event will append the remaining bytes to sd_inbuf and parsing will succeed on the complete header.

Fixes #667


apps/restconf/restconf_native.c
```diff
@@ -705,13 +705,33 @@ restconf_http1_process(restconf_conn *rc,
...
             clixon_err(OE_UNIX, errno, "cbuf_append");
             goto done;
         }
+        /* Wait for full header before parsing (#667). Only delay when the
+         * request line starts with a RESTCONF-accepted method; anything
+         * else (e.g. TLS handshake on an HTTP port) is left for the parser
+         * to reject with 400. Need >= 8 bytes (longest is "OPTIONS ") to
+         * classify; until then, assume incomplete and wait. */
+        {
+            const char *inbuf = cbuf_get(sd->sd_inbuf);
```
olofhagsand: Clixon convention requires all local variables declared at the top of the function, not in an inner block mid-function. The
  inner {} block was added specifically to work around this. These should be hoisted to the function's variable declarations.
See CONTRIBUTING.md

apps/restconf/restconf_native.c
```diff
@@ -705,13 +705,33 @@ restconf_http1_process(restconf_conn *rc,
...
+            int looks_http = buflen < 8 ||
+                strncmp(inbuf, "OPTIONS ", 8) == 0 ||
+                strncmp(inbuf, "HEAD ",    5) == 0 ||
+                strncmp(inbuf, "GET ",     4) == 0 ||
+                strncmp(inbuf, "POST ",    5) == 0 ||
+                strncmp(inbuf, "PUT ",     4) == 0 ||
+                strncmp(inbuf, "PATCH ",   6) == 0 ||
+                strncmp(inbuf, "DELETE ",  7) == 0;
+            if (looks_http && strstr(inbuf, "\r\n\r\n") == NULL){
+                goto ok;
```
olofhagsand: There should probably be a timer to avoid unbound waits on this session?

olofhagsand (commented):
Logic looks correct, had a minor style comment, a proposal for a timer.
Also, can one make a test for this case, ie that exercises the code?
It would mean sending a partial message, waiting, and then sending the rest, which may be non-trivial?

hoxnox:
I've added test and resolved all remarks.

olofhagsand:
There is a problem with nc vs netcat on linux vs musl/alpine vs freebsd, I had those problems before, see test/lib.sh

hoxnox:
corrected

> There is a problem with nc vs netcat on linux vs musl/alpine vs freebsd, I had those problems before, see test/lib.sh

hoxnox:
corrected tests

hoxnox:
test timing adjusted

olofhagsand:
Hey, I understand the test might be difficult to do, we could pass it if you do some ad-hoc testing for this specific thing.

hoxnox:
> Hey, I understand the test might be difficult to do, we could pass it if you do some ad-hoc testing for this specific thing.

All tests is green. Everything is ready to merge.

olofhagsand:
My bad, I was looking at an earlier run

hoxnox:
Is there anything else I need to do for that PR?

olofhagsand:
I will look at it soon

apps/restconf/restconf_native.c
```diff
@@ -730,8 +737,13 @@ restconf_http1_process(restconf_conn *rc,
             strncmp(inbuf, "PUT ",     4) == 0 ||
             strncmp(inbuf, "PATCH ",   6) == 0 ||
             strncmp(inbuf, "DELETE ",  7) == 0;
-        if (looks_http && strstr(inbuf, "\r\n\r\n") == NULL)
+        if (looks_http && strstr(inbuf, "\r\n\r\n") == NULL){
```
olofhagsand: strstr works on null-terminated strings, but inbuf may not be, use memmem instead

apps/restconf/restconf_native.c
```diff
@@ -1088,6 +1161,8 @@ restconf_connection_close1(restconf_conn *rc)
     }
     rsock = rc->rc_socket;
     clixon_debug(CLIXON_DBG_RESTCONF, "%s", rsock->rs_description?rsock->rs_description:"");
+    if (restconf_header_timer_unreg(rc) < 0)
```
olofhagsand: May leak an open fd, place after close (or do close even if unreg fails)

olofhagsand (changes requested):
Consider squashing the history into a single commit
(Apologies for the delay)

hoxnox:
done

olofhagsand:
Thanks, I noticed memmem is non-posix, sorry, I will merge it and fix that later,

PR #680, opened 2026-06-22
Sort RESTCONF payload tree before optimized XPath evaluation. Fixed issue 679

devibinoy:
Optimized XPath lookup may operate on a freshly parsed PATCH payload before its children are sorted according to YANG ordering.

Sort the temporary payload tree before evaluating XPath expressions to ensure optimized list lookups work correctly.


olofhagsand:
Pushed a slight modification patch instead, including a regression test,
closing this

PR #682, opened 2026-07-05
CLI show compare via RFC 9144 NETCONF compare RPC

joc-se:
The cli client makes an RPC to the netconf compare method, RFC9144, which returns a yang patch. The client traverses all edits in the received patch and creates a list of all differences.
Note that the cli command is the same as before, but the result has a slightly different format.
Only xml format is currently supported, other formats are handled by the client as before.


joc-se:
The output from the command now clearly shows where in the configuration differences exist, see the example below.

Previous output format from cli commad:

```
<multi xmlns="urn:example:clixon">
-            <parameter>
-               <first>a1</first>
-               <second>a2</second>
-               <value>17</value>
-               <value>18</value>
-            </parameter>
+            <parameter>
+               <first>c1</first>
+               <second>c2</second>
+               <value>72</value>
+               <value>73</value>
+            </parameter>
            <parameter xmlns="urn:example:clixon">
               <first>d1</first>
               <second>d2</second>
+               <value>97</value>
-               <value>99</value>
            </parameter>
         </multi>

```
New output format from cli command:

```
--- running
+++ candidate
/clixon-example:top/section=y/multi/parameter=a1,a2
-   <parameter xmlns="urn:example:clixon">
-      <first>a1</first>
-      <second>a2</second>
-      <value>17</value>
-      <value>18</value>
-   </parameter>
/clixon-example:top/section=y/multi/parameter=c1,c2
+   <parameter xmlns="urn:example:clixon">
+      <first>c1</first>
+      <second>c2</second>
+      <value>72</value>
+      <value>73</value>
+   </parameter>
/clixon-example:top/section=y/multi/parameter=d1,d2/value=97
+   <value xmlns="urn:example:clixon">97</value>
/clixon-example:top/section=y/multi/parameter=d1,d2/value=99
-   <value xmlns="urn:example:clixon">99</value>

```

lib/src/clixon_xml_io.c
```diff
@@ -1293,450 +1290,6 @@ clixon_xml_attr_copy(cxobj      *xin,
...
- *    if (clixon_xml_diff2cbuf(cb, 0, x0, x1) < 0)
- *       err();
- *    cligen_output(stdout, "%s", cbuf_get(cb));
- * @endcode
- * @see xml_diff which returns diff sets
- * @see clixon_compare_xmls which uses files and is independent of YANG
- * @see clixon_text_diff2cbuf
- */
-int
-clixon_xml_diff2cbuf(cbuf  *cb,
```
olofhagsand, line 1733: Actually this is used in the controller for now: see https://github.com/clicon/clixon-controller/blob/82c846c6492dcb358358507e2bba12ea65beb043/src/controller_rpc.c#L2694
I expect to replace this code as well eventually, but for now the fn needs to stay in place (possibly within an ifdef?)

apps/cli/cli_common.c
```diff
@@ -903,6 +926,8 @@ compare_db_names(clixon_handle    h,
         xml_free(xc1);
     if (xc2)
         xml_free(xc2);
+    if (xret)
+        xml_free(xret);
     return retval;
 }
```
olofhagsand, line 933: Style: please remove extra LF

lib/src/clixon_proto_client.c
```diff
@@ -2495,3 +2495,64 @@ clixon_rpc_nacm_autocli_filter(clixon_handle          h,
...
+    retval = 0;
+  done:
+    clixon_debug(CLIXON_DBG_DEFAULT | CLIXON_DBG_DETAIL, "retval:%d", retval);
+    if (cb)
+        cbuf_free(cb);
+    if (xret)
+        xml_free(xret);
+    return retval;
+}
+
```
olofhagsand, line 2558: Style: rm extra LF

test/test_cli_diff.sh
```diff
@@ -166,8 +169,7 @@ new "check candidate"
 expectpart "$($clixon_cli -1 -f $cfg show config candidate)" 0 "^<top xmlns=\"urn:example:clixon\"><section><name>x</name><table><parameter><name>b</name><value>42</value></parameter><parameter><name>c</name><value>72</value></parameter><parameter><name>d</name><value>99</value></parameter></table></section></top>$"
 
 new "check compare xml b"
-echo "$clixon_cli -1 -f $cfg show compare xml"
-expectpart "$($clixon_cli -1 -f $cfg show compare xml)" 0 "<table xmlns=\"urn:example:clixon\">" "^\-\ *<parameter>" "^+\ *<parameter>" "^\-\ *<name>a</name>" "^+\ *<name>c</name>" --not-- "^+\ *<name>a</name>" "^\-\ *<name>c</name>"
+expectpart "$($clixon_cli -1 -f $cfg show compare xml)" 0 "^/clixon-example:top/section=x/table/parameter=a" "^\-\ *<parameter xmlns=\"urn:example:clixon\">" "^\-\ *<name>a</name>" "^\-\ *<value>17</value>" "^\-\ *</parameter>" "^/clixon-example:top/section=x/table/parameter=c" "^\+\ *<parameter xmlns=\"urn:example:clixon\">" "^\+\ *<name>c</name>" "^\+\ *<value>72</value>" "^\+\ *</parameter>" "^/clixon-example:top/section=x/table/parameter=d/value" "\-\ *<value xmlns=\"urn:example:clixon\">98</value>" "^\+\ *<value xmlns=\"urn:example:clixon\">99</value>"
```
olofhagsand, line 172: It is nice with an api-path, but after checking with some downstream users, I think we need to keep the context of the change, maybe as an option?
That is, the old format was (with context):
```
<top xmlns="urn:example:clixon">
   <section>
      <name>x</name>
      <table>
         <parameter>
            <name>a</name>
-           <value>17</value>
+           <value>42</value>
         </parameter>
      </table>
```
And the new format (context replaced with api-path):
```
--- running
+++ candidate
/clixon-example:top/section=x/table/parameter=a/value
-   <value xmlns="urn:example:clixon">17</value>
+   <value xmlns="urn:example:clixon">42</value>
```

olofhagsand (commented):
Will pull this into branch "show-compare"

joc-se:
commits have been added manually to the branch show-compare

PR #686, opened 2026-08-01
Fixed crash on show xpath with non-valid namespace prefix

krihal:
Fixed crash on show xpath with non-valid namespace prefix.


olofhagsand (approved):
Looks good

Issue #684, opened 2026-07-28
Empty enum/bit description causes autocli CLI-spec generation to fail with "syntax error: at or before: '"'"


paul-sirin:
## Describe the bug

When a YANG `enum` (or `bit`) statement has an **empty** `description ""`, autocli generation for any module containing a leaf of that enumeration type fails with a CLIgen parse error:

```
Autocli for module <mod> in <file>.yang:<N>: Error: syntax error: at or before: '"'
```

## Root cause

In `lib/src/clixon_autocli_generate.c`, `yang2cli_leaf_var_sub()` emits each enum value's description as CLIgen per-choice help text:

```c
if ((yd = yang_find(yi, Y_DESCRIPTION, NULL)) != NULL &&
    (desc = yang_argument_get(yd)) != NULL){
    dlen = strlen(desc);
    cprintf(cb, "(\"");
    for (j=0; j<(int)dlen; j++){
        ...
    }
    cprintf(cb, "\")");
}
```

There is no check for `dlen == 0`. When the description is the empty string, this emits the literal text `("")` into the generated CLIgen spec.

CLIgen's grammar (`cligen_parse.y`) does not allow an empty `choicehelp`:

```
choicehelp : /* empty */ { $$ = NULL; }
           | '(' DQ charseq DQ ')' { $$ = $3; }
           ;

charseq    : charseq CHARS
           | CHARS
           ;
```

`charseq` requires at least one `CHARS` token, so `("")` (an empty string between the quotes) has no valid production and the parser fails right at the closing `"`. Note this differs from top-level command help text, which *does* have an explicit empty alternative (`cmd PDQ DQP`), so a leaf/container with an empty top-level `description` works fine — only the per-enum/per-bit `choicehelp` path lacks this.

## Steps to reproduce

Minimal YANG:

```yang
module repro {
  namespace "urn:repro";
  prefix repro;

  typedef my-enum {
    type enumeration {
      enum a {
        description "ok value";
      }
      enum b {
        description "";
      }
    }
  }

  container top {
    leaf x {
      type my-enum;
    }
  }
}
```

Loading this module in a clixon CLI backend with autocli enabled triggers the syntax error during CLI-spec generation for `top`/`x`.

## Expected behavior

An empty enum/bit description should either be treated the same as "no description" (i.e. no help text is emitted for that choice) or the empty description should still be quoted in a form the CLIgen grammar can parse (e.g. skip emitting the `("...")` wrapper entirely when `dlen == 0`).

[...]

PR #689, opened 2026-08-04
Fixed: [Empty enum/bit description causes autocli CLI-spec generation to fail with "syntax error: at or before: '"'"](https://github.com/clicon/clixon/issues/684)

krihal:


test/test_autocli_enum_helpstr.sh
```diff
@@ -3,6 +3,8 @@
 # CLI help string for enum values should show per-enum description,
 # not the enclosing leaf description.
 # Tests both inline enum and typedef enum.
+# Also regression test for https://github.com/clicon/clixon/issues/684
+# An empty enum/bit description must not generate ("") which CLIgen cannot parse.
```
olofhagsand, line 7: I added a patch in cligen allowing it to parse "" for choice.
See: https://github.com/clicon/cligen/commit/302fa7d297b8bfe46648538667d022bb8ffc4cd3
The tests may still be relevant but not the fix since the desc can be strlen = 0

olofhagsand (changes requested):

PR #692, opened 2026-08-22
Fixed: NACM fails for devices using different revisions of the same YANG model

krihal:
Fixed: NACM fails for devices using different revisions of the same YANG model (https://github.com/clicon/clixon-controller/issues/251).

### DIFF
--- a/example/main/example_backend.c
+++ b/example/main/example_backend.c
@@ -38,6 +38,7 @@
   *  -a <..> Register callback for this yang action
   *  -m <yang> Mount this yang on mountpoint
   *  -M <namespace> Namespace of mountpoint, note both -m and -M must exist
+  *  -R  Use the key of the list enclosing the mountpoint as revision of the mounted module (requires -m and -M)
   *  -n  Notification streams example
   *  -r  enable the reset function
   *  -s  enable the state function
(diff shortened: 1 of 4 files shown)


example/main/example_restconf.c
```diff
@@ -400,6 +432,8 @@ example_restconf_yang_mount(clixon_handle   h,
         /* In yang name+namespace is mandatory, but not revision */
         cprintf(cb, "<name>%s</name>", _mount_yang); // mandatory
         cprintf(cb, "<namespace>%s</namespace>", _mount_namespace); // mandatory
+        if (_mount_revision_from_key && (revision = mount_point_key(xt)) != NULL)
+            cprintf(cb, "<revision>%s</revision>", revision);
```
olofhagsand: Use xml_chardata_cbuf_append(cb, 0, revision) instead of a direct %s for encoding, to avoid key injections.

example/main/example_backend.c
```diff
@@ -1124,6 +1158,8 @@ main_yang_mount(clixon_handle   h,
         /* In yang name+namespace is mandatory, but not revision */
         cprintf(cb, "<name>%s</name>", _mount_yang); // mandatory
         cprintf(cb, "<namespace>%s</namespace>", _mount_namespace); // mandatory
+        if (_mount_revision_from_key && (revision = mount_point_key(xt)) != NULL)
+            cprintf(cb, "<revision>%s</revision>", revision);
```
olofhagsand: Use xml_chardata_cbuf_append(cb, 0, revision) instead of a direct %s for encoding, to avoid key injections, that is:
```
   cprintf(cb, "<revision>");
   xml_chardata_cbuf_append(cb, 0, revision);
   cprintf(cb, "</revision>");
```
It might not matter in the real system since this is just example code, but its still good to keep this as standard.

example/main/example_backend.c
```diff
@@ -73,7 +73,7 @@
 #include <clixon/clixon_backend.h>
 
 /* Command line options to be passed to getopt(3) */
-#define BACKEND_EXAMPLE_OPTS "a:m:M:n:o:O:rsS:x:iuUtV:"
+#define BACKEND_EXAMPLE_OPTS "a:m:M:Rn:o:O:rsS:x:iuUtV:"
```
olofhagsand, line 77: Also add -R in the summary on lines 38-51

krihal:
Requested changes added.

PR #694, opened 2026-09-03
cli: persist history after every command, not only at exit

khromenokroman:
@olofhagsand Hi
Previously cli_history_save() was called once, from cli_terminate(), so the CLI history file was only ever written when the process exited cleanly. If the CLI session was killed, crashed, or lost connection, none of the commands entered in that session were saved.

Flush history right after each command is read in cli_interactive(), using the same cli_history_save()/cligen_hist_file_save() already used at exit. This is cheap since the history ring buffer is small (CLIGEN_HISTSIZE_DEFAULT / CLICON_CLI_HIST_SIZE, 100 lines by default).

### DIFF
--- a/apps/cli/cli_main.c
+++ b/apps/cli/cli_main.c
@@ -75,6 +75,8 @@
 #define CLI_OPTS "+hVD:f:E:l:C:F:1sa:u:d:m:qp:GLy:c:U:g:o:"
 
 static char *_restarg = NULL; /* what remains after options XXX just to avoid mem-leakage, try to fix wo global var */
+static long  _hist_offset = 0; /* byte offset up to which the CLI history file has
+                                 * been read into this session (see cli_history_sync) */
 
 /*! Check if there is a CLI history file and if so dump the CLI histiry to it
  *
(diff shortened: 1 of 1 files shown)


v1shnya:
@olofhagsand hi! 
it could help us, please, consider!

olofhagsand:
I am not sure, the current history behavior was modeled after how a unix shell (bash) behaves by default.
In particular, in the case of multiple CLI:s in parallel the history would be mixed between different sessions, whereas in the current solution the would follow the sessions linearly (but as you say only after their termination).
Do you have other examples of tools/clis which do as you propose?
I am not saying "no", just wanted to invite to a discussion and further motivation.

khromenokroman:
@olofhagsand 
  Thanks for the feedback — the concern about concurrent sessions was valid, and here's a concrete example you asked for: zsh's 
  SHARE_HISTORY option does exactly this kind of real-time history sharing across parallel interactive sessions, and it avoids the
  clobbering problem by never rewriting the history file. I reworked the patch to follow the same model.

  Original approach (what you flagged): flush on every Enter by calling the existing cli_history_save(), which opens the file "w+" and
  rewrites it entirely from the in-memory ring buffer. With two sessions both flushing on every keystroke, whichever session flushes last
  wins and silently discards whatever the other had just written.
  Revised approach, in a new cli_history_sync() called after every command instead of cli_history_save():
  1. Track a byte offset per session (starts at EOF after the initial load in cli_history_load()).
  2. On each Enter: fseek() to that offset and read only the bytes appended since — i.e. whatever other sessions have written in the
     meantime — and merge them into this session's in-memory history via the existing cligen_hist_file_load().
  3. Append ("a", not "w+") only the command just entered. The file is never truncated in this path.

  So two sessions interleaving commands end up with all of them in the file, in the order actually written, instead of one session's
  history disappearing. I tested this directly with a small harness that interleaves two in-process "sessions" writing to the same file
  and confirmed nothing is lost or reordered incorrectly.

  The existing cli_history_save() (full rewrite) is untouched and still runs once, at exit — that's the same behavior clixon already had
  before this change, and it's what trims the on-disk file down to the configured history size (CLICON_CLI_HIST_SIZE).

  One caveat worth flagging: there's still a small window between the read and the append within a single cli_history_sync() call where
  another session could write in between — same class of race zsh itself has without file locking. Given the small window and low
  frequency of truly simultaneous keystrokes across sessions, I think that's an acceptable tradeoff, but happy to discuss if you'd rather
  see a lock around the file.

  Let me know if this addresses your concern and I'll push the updated commit.

v1shnya:
@olofhagsand Hi Olof,

Yes, let’s first clarify what we want to achieve.

According to the Clixon documentation on persistent command history:
“Only the latest session dumping its history will survive (bash merges multiple session history).”

So, if the same user has multiple CLI sessions, it looks like not all commands from all sessions will be saved to the history file.

We should also keep in mind that Clixon is a single-threaded application. If a plugin crashes and the whole process has to restart, it could be useful to have a complete history of all previously entered commands, stored immediately and in chronological order.

In Linux/bash, saving each entered command to the history file has a simple and well-proven solution:
https://unix.stackexchange.com/questions/131504/how-to-sync-terminal-session-command-history-in-bash 

To avoid changing the current behavior, I would propose introducing a new option, for example CLICON_CLI_HIST_MODE, which could switch between:
- the current history-saving behavior;
- saving each command to the history file immediately after it is entered.

If you agree with the proposal we could make a PR for your review!
 
Best regards,
Andrey

